PR #4, opened 2016-06-16
Fix wrong regular expression

PrestanceDesign:
The . character in a regular expression means "any character", so all scripts are blocked.


julienbourdeau:
Thank you for your contribution to the doc !

PR #16, opened 2016-08-08
Markdown corrections for anchor link and embedded code.

inetbiz:
Changes to `/themes/getting-started/setup-env.rst`

### DIFF
--- a/themes/getting-started/setup-env.rst
+++ b/themes/getting-started/setup-env.rst
@@ -1,14 +1,10 @@
-****************
-Setting up your local environment
-****************
-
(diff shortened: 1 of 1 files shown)


xBorderie:
Thank you for this, but the new doc files are not written using Markdown, but [reStructured Text](http://docutils.sourceforge.net/rst.html). I know, it's painful, but that's what [Sphinx](http://www.sphinx-doc.org/en/stable/) uses.

inetbiz:
ok. I'll study up on that! I had been using an app called MarkdownPad2  http://markdownpad.com/

PR #20, opened 2016-09-09
Introduce Sphinx and update doc

julienbourdeau:
Here is what it looks like so far:

helpers__functions_and_modifier_ _prestashop_developer_documentation_1_7_0_0_documentation


xBorderie:
:heart_eyes:

julienbourdeau:
I merged since it's more a starting point from where we can start adding/editing doc.

PR #30, opened 2016-10-20
Use proper fetch method for template call in renderWidget() example

prestamodule:
Follow this commit: https://github.com/PrestaShop/ps_contactinfo/commit/293b5202229034971a458cca6fe2d48b1730bfbb


xBorderie:
Thanks @prestamodule!

julienbourdeau:
Thank you @prestamodule !

PR #50, opened 2017-01-10
Update info about Bootstrap.

xBorderie:


julienbourdeau:
👍

prestarocket:
The theme is updated with TB v4 alpha 6 ?

maximebiloe:
Hum... No. This is the purpose of this PR, we wanted to clarify.

xBorderie:
@prestarocket We recommend to use Bootstrap 4 alpha 4 rather than any version of Bootstrap 4, as it was written before.

prestarocket:
you plan to do the same mistake with PS 1.6 : use an old version of TB and not be able to use last version ?

kpodemski:
Yet another day in PrestaShop community.

Using alpha was not a great idea in the first place, using outdated framework at the beginning of 1.7 life is yet another one. You should update this now when not many people are using 1.7.

PR #53, opened 2017-02-14
Update theme-yml.rst

dreamtheme:
Syntax for hook exceptions do NOT work! there is a bug report on the forge about it. Fix it or remove it from the documantation as it does not work.


xBorderie:
It's been fixed in v1.7.1.2: http://forge.prestashop.com/browse/BOOM-2155
Thank you!

PR #54, opened 2017-02-15
creating module: fix setTemplate path

MaciejKucia:
Templating behaviour has changed since 1.7
Current docs are **very** confusing


maximebiloe:
Thank you @MaciejKucia for noticed

PR #58, opened 2017-09-23
Removed some spaces in folder and file paths

JoshuaLuckers:
Removed some spaces in folder and file paths that could cause some annoyances when copying folder and file paths from the table with possible files and folders for a PrestaShop 1.6 module.


xBorderie:
Well spotted, thanks!

PR #86, opened 2018-05-03
Add page about overrides

Quetzacoalt91:
Summary:

* Why do they exist
* Important note about override limitations & dangers
  * hooks
  * Controllers
  * Contributions welcomed
* How to override classes & controllers
* How to override theme templates

### DIFF
--- a/src/content/1.7/modules/concepts/overrides.md
+++ b/src/content/1.7/modules/concepts/overrides.md
@@ -39,13 +39,13 @@ For instance you could have a single Tools class for core and module helper meth
 ```php
 class MyModuleTools extends Tools
 {
-  /**
-   * Adding a new method for the example
-   */
-  pulic static function array_pluck(array $data, 'property')
-  {
-    // [...]
-  }
+    /**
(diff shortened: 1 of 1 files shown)


src/content/1.7/modules/concepts/overrides.md
```diff
@@ -0,0 +1,138 @@
...
+
+Let's consider the file `/modules/<module_name>/override/classes/controllers/FrontController.php`, overriding only one method of the core file.
+
+```php
+/*
+ * With this override, you have a new Smarty variable called "currentController" available in header.tpl
+ * This allows you to use a different header if you are on a product page, category page or home.
+ */
+class FrontController extends FrontControllerCore {
+    public function initHeader()
```
PierreRambaud: Use 4 spaces everywhere, or 2 spaces to be more consistent.

PierreRambaud: :+1:

mickaelandrieu:
Thank you @Quetzacoalt91 !

PR #88, opened 2018-05-03
Linux installation

toutantic:
Add a script to download hugo binary.

We will need to remove the binary from the repository history using bfg-repo-cleaner
https://help.github.com/articles/removing-sensitive-data-from-a-repository/


Quetzacoalt91:
It seems we can also use the npm package: https://www.npmjs.com/package/hugo-bin

toutantic:
I think I hate npm packages
Installating hugo for our specific use case:
https://github.com/toutantic/docs/blob/linuxInstallation/bin/installHugo.sh

Part of the nodejs code we don't need at all:
https://github.com/fenneclab/hugo-bin/blob/master/lib/index.js

If we want a clean install of hugo just follow the documentation: https://gohugo.io/getting-started/installing/

toutantic:
How do I test if my changes work on travis without merging ??

Quetzacoalt91:
You have to enable it on your fork: https://travis-ci.org/toutantic/docs.

Then push again to trigger the event.

Quetzacoalt91:
Hmm, do you really think we don't need that js part?
It seems to handle all the main operating systems, this would allow to go extend this PR perimeter.

README.md
```diff
@@ -8,7 +8,7 @@ Install guide: https://gohugo.io/getting-started/installing
 
 TL;DR:
 
-* **Linux**: Use the binary in the `/bin` directory
+* **Linux**: `./bin/installHugo.sh`
```
mickaelandrieu: :+1:

mickaelandrieu:
@toutantic I'm in favor of merging this, what are the risks?

eternoendless:
The risk is breaking the build and putting the whole doc offline (there's an `rm -rf public` at the start of the deployment script)

We can revert it if something fails, but it would be best to adapt the travis script so that we can test PRs.

mickaelandrieu:
Sadly, this needs to be rebased... wdyt @eternoendless?

eternoendless:
Travis now tests PRs without deploying them, if you rebase we can see how long it takes.

bin/installHugo.sh
```diff
@@ -0,0 +1,6 @@
+#/bin/bash
+set -euo pipefail
```
mickaelandrieu: I have an error with this option: 
```
➜  ps-docs git:(master) ✗ ./installHugo.sh
./installHugo.sh: 2: set: Illegal option -o pipefail
```

PierreRambaud: Are you sure you're running with bash, and not sh?
Because I can only reproduce under sh context, not even on Bash, Zsh, or anything else :/

PierreRambaud: But you locate something false, the shebang isn't good:
`#/bin/bash` must be `#!/bin/bash` That's maybe why you have some problem. (Your configuration maybe use sh by default)

mickaelandrieu: So the shebang needs to be updated, am I right?

PierreRambaud: Yep!

eternoendless:
Thank you @toutantic

PR #90, opened 2018-05-04
Separate core & 3rd party module translation implementation

Quetzacoalt91:
On prestaShop 1.7, the translation system has changed for native modules, but 3rd party modules must still use the previous one.

To avoid confusion, I moved the native details in the development chapter, and used the legacy one in the module chapter.

### DIFF
--- a/src/content/1.7/development/native_module_translation.md
+++ b/src/content/1.7/development/native_module_translation.md
@@ -1,5 +1,5 @@
 ---
-title: Module translation
+title: Native module translation
 weight: 7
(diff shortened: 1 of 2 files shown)


PierreRambaud:
Just a remark about the php spaces, maybe use 4 spaces like other pages. :smile:

Quetzacoalt91:
Thanks, I have just taken your point in account.

mickaelandrieu:
Thank you @Quetzacoalt91

PR #95, opened 2018-05-31
add benchmarks section

jocel1:


mickaelandrieu:
Thanks for the tutorial, I'm a little bit busy but I'll come back to review asap 👍

toutantic:
Fantastic !!
I think this page should be split in two:
- How to benchmark your PrestaShop shop
- How to Optimize the performance of your PrestaShop
They should of course cross-reference each other.

And if we extract the chapter ## Optimize your LAMP stack I propose the following table of content:

# How to benchmark your PrestaShop shop
## Prerequisite
## Benchmark methodology

## Prepare your benchmark
### Prepare your dataset
#### How to use this dataset during Prestashop install?
### Monitor your server 
### Prepare your shop

## Bench your shop performances
### Record your settings
Server configuration (CPU/mem/Disk/...)
PHP/Apache/Mysql settings
PrestaShop configuration
Dataset used
### Run the benchmark
### Read the result

jocel1:
ok no problem, I'll change the structure :)

PR #99, opened 2018-06-26
 replace "mod" parameter by "d" in translation example for 1.7

kermorgant:


Quetzacoalt91:
Thank you @kermorgant

PR #100, opened 2018-06-26
Update code in adding_configuration_page.md

Xikren:
The return statement in getContent() is not reached if isSubmit() is false and displayForm() doesn't return anything.


Xikren:
I'm learning to use PrestaShop and the code in the guided tutorial for mudules on PrestaShop DevDocs have some flaws.

Quetzacoalt91:
Thank you @Xikren

PR #102, opened 2018-07-02
Improve seo

PierreRambaud:
We need to use hyphen instead of underscore for urls, name and everything when you separate words:
https://blog.codinghorror.com/of-spaces-underscores-and-dashes/
(many other articles available on the Internet, 99% of them are claiming for hyphen / dashes)

More, add redirect from old content to the new one.


eternoendless:
Thank you @PierreRambaud

PR #104, opened 2018-07-10
typo fix

hellotaotao:


PierreRambaud:
Nice catch @hellotaotao Thanks ;)

PR #113, opened 2018-08-08
Fix event name "reponsive update"

dennispw:
Looking at this list: https://devdocs.prestashop.com/1.7/themes/reference/javascript-events/#dispatched-events

The following event is wrong:
`responsiveUpdate`

Reading the documentation one would assume to use it as such:
`prestashop.on("responsiveUpdate", ... );`

White the actual, currect event name and usage is:
`prestashop.on("responsive update", ... );`

### DIFF
--- a/src/content/1.7/themes/reference/javascript-events/_index.md
+++ b/src/content/1.7/themes/reference/javascript-events/_index.md
@@ -58,7 +58,7 @@ Event Name            | Description
  `updateProduct`      | On the product page, selecting a new combination will reload the DOM via ajax calls. After the update, this event is fired.
  `handleError`        | This event is fired after a fail of POST request. Have the `eventType` as first parameter.
  `updateFaces`        | On every product list page (category, search results, pricedrop and so on), the list is updated via ajax calls if you change filters or sorting options. Each time the facets is reloaded, this event is triggered.
- `responsiveUpdate`   | While broswer is resized, this event is fired with a `mobile` parameter.
+ `responsive update`   | While broswer is resized, this event is fired with a `mobile` parameter.
 
 ### Triggering delegated events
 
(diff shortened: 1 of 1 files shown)


dennispw:
Yes, comparing it to the others it should definately be `responsiveUpdate` so that everything follows the same naming convention and camelCase but I guess it slipped through the cracks somewhere.

mickaelandrieu:
Hello @dennispw,

do you know how we should declare it? We need to keep the old event name for compatibility but we may add the *responsiveUpdate* too in the next release so you may rely on it in the future, wdyt?

dennispw:
@mickaelandrieu I'm not sure since I haven't located where it's actually defined. But I guess it _should_ only need to be changed from `responsive update` to `responsiveUpdate` (and revert this PR I guess). If people are using `responsive update` it might be a good idea to add some sort of alias or additional event name so that both `responsive update` and `responsiveUpdate` works. But then again, the cleanest solution in my opinion would be to make it so that `responsiveUpdate` (as of right now) is the same in the documenation and in the actual code and usage. Of course you need to be really clear about this change so that people can update their themes as soon as possible.

In terms of the actual name, since it's really only being fired when below/above a specific width I guess the name should be something like `breakpointUpdate`, `breakpointState` or something similar. If using a more "generic" name it might be easier to add to it without worrying if it makes sense together with the eventname.

PierreRambaud:
```
themes/classic/_dev/js/components/top-menu.js
63:    prestashop.on('responsive update', function(event) {

themes/classic/_dev/js/responsive.js
58:	prestashop.emit('responsive update', {
```
And was added here https://github.com/PrestaShop/PrestaShop/commit/9fc642b16323c8ac98d2cbbd8a964ca381acb33c#diff-e3e9e4751854d40668ab65e66cb22cf1R40 :)
We can merge this pull request, and try to make something better in the classic theme :)

PR #118, opened 2018-08-23
Fix code example for setTemplate in controller

alu-:


mickaelandrieu:
the 2 forms should works but I think it's better to document only one way to do it :+1:

mickaelandrieu:
Thanks @alu-

PR #122, opened 2018-08-24
Update hooks.md

jolelievre:
use `prestashop.core.hook.dispatcher` instead of `prestashop.hook.dispatcher`


mickaelandrieu:
Thanks @jolelievre

PR #134, opened 2018-09-10
Update grid tutorial

jolelievre:
Change grid services yaml definition with updated arguments, reorganize the service order to focus on what needs to be built, updates the controller code and provide information about the Filters auto resolved class.


mickaelandrieu:
Thanks @jolelievre

PR #137, opened 2018-09-12
duplicate word

vhessam:


PierreRambaud:
Thanks @vhessam nice catch ;)

PR #140, opened 2018-09-17
Correct a typo

ptt-homme:


PierreRambaud:
Nice catch! Thanks @ptt-homme

PR #161, opened 2018-11-07
Added more information on PHP configuration

mickaelandrieu:


src/content/1.7/scale/optimizations.md
```diff
@@ -20,15 +20,61 @@ of concurrent users you want to simulate during the bench.
...
+date.timezone = UTC
+
+[Session]
+session.auto_start = Off
+
+[PHP]
+short_open_tag = Off
+realpath_cache_size = 4096K
+realpath_cache_ttl = 600
+apc.enable_cli = 1
```
PierreRambaud: useless and apc is not mandatory or even installed by default :)

src/content/1.7/scale/optimizations.md
```diff
@@ -20,15 +20,61 @@ of concurrent users you want to simulate during the bench.
...
+display_errors = Off
+
+magic_quotes_gpc = off
+memory_limit = 512M
```
PierreRambaud: add a comment because you can set more memory if you have only one server with more memory than 2Gb :)

src/content/1.7/scale/optimizations.md
mickaelandrieu: why is it useless?

PierreRambaud: Because it's only for cli, prestashop is not running in cli. 
Never mix cli and httpd configuration it's a bad practice.

src/content/1.7/scale/optimizations.md
```diff
@@ -20,15 +20,61 @@ of concurrent users you want to simulate during the bench.
...
+max_execution_time = 300
+max_input_time = 300
+upload_max_filesize = 20M
```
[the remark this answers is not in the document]
mickaelandrieu: so, what update do you expect from me?

mickaelandrieu:
> just add 2M of post_max_size, otherwise could encounter some problems :D

Sir! Yes Sir! 👍

PierreRambaud:
Thanks @mickaelandrieu for your work :)

PR #164, opened 2018-11-20
Translations in other classes must use class name

rstefan:
Calling translation method from other classes that aren't module class or controller class must use a second parameter to fetch the corresponding translation text.
I add a sample of code to explain the call.

Regards.


mickaelandrieu:
Thanks @rstefan !

PR #167, opened 2018-11-27
Update use-hooks-on-modern-pages.md

DguillonStark:
After many search, i had to use this " require_once _PS_MODULE_DIR_.'foo/src/Repository/ProductRepository.php';
 " in my foo.php, to avoid Uncaught PHP Exception Symfony\Component\Debug\Exception\ClassNotFoundException.

And i saw many people asking for it without answer.
So may be this isn't the best solution, but something is missing on this page to make this module work properly.

### DIFF
--- a/src/content/1.7/modules/concepts/hooks/use-hooks-on-modern-pages.md
+++ b/src/content/1.7/modules/concepts/hooks/use-hooks-on-modern-pages.md
@@ -147,6 +147,14 @@ You can now use it in your module (and everywhere in PrestaShop modern pages!):
 ```php
 // foo.php
 
+if (!defined('_PS_VERSION_')) {
+    exit;
+}
+
+// Add this 
+require_once _PS_MODULE_DIR_.'foo/src/Repository/ProductRepository.php';
+
+
 /* ... */
 /**
  * Get the list of products for a specific lang.


matks:
@DguillonStark could you please tell us where you saw people asking for this ?

@sarjon is right: the right way is to use composer autoload which really makes a php developer life's easier 😄 but we can improve the way we document it and spread the info

DguillonStark:
Hi,

@matks, i saw it on the prestashop forum.
[https://www.prestashop.com/forums/topic/921955-uncaught-php-exception-symfonycomponentdebugexceptionclassnotfoundexception-help/?page=0#comment-2999611](url)
[https://www.prestashop.com/forums/topic/914002-cr%C3%A9ation-de-services-et-erreur-namespace-pour-un-module/](url)
[https://www.prestashop.com/forums/topic/903916-symfony-service-registeration-error/?page=0#comment-2972364](url)

Each of them follows this tutorial.

I've never had to use composer before, maybe an example of using  composer will be fine for this page, since we don't use  composer before prestashop 1.7 :)

matks:
@DguillonStark here is a nice tutorial about how to use Composer autoloading: https://phpenthusiast.com/blog/how-to-autoload-with-composer 😉 

Composer is a great php tool that provides you 2 major features:
- it is a dependency manager: this means if you use external libraries it provides a robust way to download and update them and it ensures they are compatible between themselves
- it provides a very useful autoloading (this is what you are looking for): by writing the right composer.json file and following some conventions, you can stop using `require` and `include` php keywords 😉 you do not need them anymore

PR #168, opened 2018-11-28
Add a paragraph in 1.7.5 changes page about the _legacy_link

jolelievre:


src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -45,6 +45,33 @@ public function hookdisplayFooterProduct($params)
 }
 ```
 
+## About getAdminLink and _legacy_link
+
+The 1.7.5 version offers a conversion from legacy links to new migrated Symfony pages. Meaning if you use the `Link::getAdminLink`
```
PierreRambaud: no need to break line, let markdown do it alone :)

jolelievre: I know it was juste a matter of readability, but I will write it on one line

PierreRambaud: Your IDE should do it alone. (Let's come to Emacs \o/)
https://github.com/PrestaShop/docs/blob/462b6f79b318876200d5986cc54431e5ceb71a11/src/content/1.7/modules/core_updates/1.7.5.md#about-getadminlink-and-_legacy_link
It is weird in GitHub

jolelievre: You mean the last part with `[controller-routing]: {{< relref ...`? Yep in github it does
But I tested it with hugo and it works fine It's used to define the link used before in `[Controller and Routing][controller-routing] page`

src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -45,6 +45,28 @@ public function hookdisplayFooterProduct($params)
...
+
+```php
+    //You shouldn't write
+    $link->getAdminLink('AdminOrders') . '&amp;id_order={$order->id|intval}&amp;vieworder';
+    
+    //But rather
+    $link->getAdminLink('AdminOrders', true, [], ['id_customer' => $customer->id|intval, 'viewcustomer' => 1]);
+```
+
+Although this won't be a problem for legacy controllers, but as they will get migrated getAdminLink won't work any more if it is not used properly, so we advice you to refactor your getAdminLink beforehand.
```
Quetzacoalt91:
```suggestion
Although this won't be a problem for legacy controllers, the generated link won't work any more if it is not used properly once migrated to Symfony. So we advice you to refactor your call to `getAdminLink` beforehand.
```

src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -45,6 +45,28 @@ public function hookdisplayFooterProduct($params)
...
+
+You can find more details about this new feature in the [Controller and Routing][controller-routing] page. 
+
 **Links:**
 
-* [Related issue on module Google Ananlytics (GitHub)](https://github.com/PrestaShop/ps_googleanalytics/pull/14/files)
\ No newline at end of file
+* [Related issue on module Google Ananlytics (GitHub)](https://github.com/PrestaShop/ps_googleanalytics/pull/14/files)
```
Quetzacoalt91: I guess this link should remain with the related chapter.

jolelievre: yep, probably

src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -47,4 +47,26 @@ public function hookdisplayFooterProduct($params)
...
+This prevents module developer to update their links each time a page has been migrated. However in many places the method is not used with the url parameters as arguments. Parameters are concatenated after the getAdminLink call. This can result in wrong conversion use cases, so you might need to refactor the way you use this method.
+
+Example:
+
+```php
+    //You shouldn't write
+    $link->getAdminLink('AdminOrders') . '&amp;id_order={$order->id|intval}&amp;vieworder';
+    
+    //But rather
+    $link->getAdminLink('AdminOrders', true, [], ['id_customer' => $customer->id|intval, 'viewcustomer' => 1]);
```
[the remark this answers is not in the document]
jolelievre: Well the method already managed these arguments in 1.7.0.x version https://github.com/PrestaShop/PrestaShop/blob/32b208976e70e98080e8e31ab9e669e679822ac2/classes/Link.php#L675
But for 1.6 versions I don't know how we can do..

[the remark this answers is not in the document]
mickaelandrieu: the API in 1.6.1.x was ugly as f***!

How about suggest this instead:

```php
$link->getAdminLink('AdminOrders', true, [], ['id_customer' => $customer->id|intval, 'viewcustomer' => 1]) . '&amp;id_order={$order->id|intval}&amp;vieworder';
```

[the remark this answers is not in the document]
jolelievre: this would work in theory, but the 1.6 method only has 2 parameters The won't be an error raised id the number of arguments does not match?

mickaelandrieu:
To sum up: are the docs "correct" with the behavior in the code?

If it does, I'm in favor of the merge: WDYT?

jolelievre:
the doc show what I changed in the core 1.7 code to make getAdminLink work correctly, in this context it does work
but indeed it won't work in 1.6 so the question is what do we advise to developers who need to make module compatible in both 1.6 and 1.7
either we provide an example for each version (the one you suggested) or we update the 1.6 so that getAdminLink is able to deal with these extra parameters but we are extending the scope of 1.6 then and I'm not sure we want to do that

jolelievre:
@mickaelandrieu I tried what you suggested in a 1.6.1 version with php 7.1:

```php
link->getAdminLink('AdminOrders', true, [], ['id_customer' => $customer->id|intval, 'viewcustomer' => 1]) . '&amp;id_order={$order->id|intval}&amp;vieworder';
`
``

it seems to work (no errors nor notices), I think I'm gonna add this as a suggestion for 1.6 developers

File: src/content/1.7/modules/core_updates/1.7.5.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (eternoendless) [reply src/content/1.7/modules/core_updates/1.7.5.md]: > the 1.6 method only has 2 parameters The won't be an error raised id the number of arguments does not match?

You can call a function with more arguments than the function defines, but not the other way around: https://3v4l.org/Yop74

So this is not a problem.

**Comment by eternoendless** [comment]
@jolelievre I took the liberty of rephrasing things a little. If you are OK with the content I'm in favor of merging this PR.

**Comment by jolelievre** [comment]
@eternoendless you did right, it's clearer this way It's ok for me too
Did you see my comment though? https://github.com/PrestaShop/docs/pull/168#issuecomment-445203298
Do you think it's something we could introduce?
too late for 1.7.5? worth considering for 1.7.6? what about 1.6.1?

File: src/content/1.7/modules/core_updates/1.7.5.md [hunk 5]
```diff
@@ -47,4 +47,29 @@ public function hookdisplayFooterProduct($params)
...
+## About getAdminLink and _legacy_link
+
+The 1.7.5 version offers a conversion from legacy links to new migrated Symfony pages. Meaning if you use the `Link::getAdminLink` method somewhere in your modules to access a page that has been migrated, PrestaShop automatically converts it to the new Symfony migrated url.
+
+This prevents module developer to update their links each time a page has been migrated. However in many places the method is not used with the url parameters as arguments. Parameters are concatenated after the getAdminLink call. This can result in wrong conversion use cases, so you might need to refactor the way you use this method.
+
+Example:
+
+```php
+    //You shouldn't write
```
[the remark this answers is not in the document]
jolelievre: done!

src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -47,4 +47,29 @@ public function hookdisplayFooterProduct($params)
...
+    $link->getAdminLink('AdminOrders') . '&amp;id_order={$order->id|intval}&amp;vieworder';
+
+    //But rather
```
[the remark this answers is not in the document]
jolelievre: I put `Recommended code from PrestaShop 1.7.5` is it ok for you?

src/content/1.7/modules/core_updates/1.7.5.md
```diff
@@ -47,4 +47,29 @@ public function hookdisplayFooterProduct($params)
...
+Example:
+
+```php
+    // Code compliant from PrestaShop 1.5 to 1.7.4
+    $link->getAdminLink('AdminOrders') . '&amp;id_order={$order->id|intval}&amp;vieworder';
+
+    // Recommended code from PrestaShop 1.7.5
+    $link->getAdminLink('AdminOrders', true, [], ['id_customer' => $customer->id|intval, 'viewcustomer' => 1]);
+
+    // If you need to be compatible with both 1.6 and 1.7 versions then you can combine both writing
```
Quetzacoalt91: This method is available from PrestaShop 1.5, let's update it here as well.
```suggestion
    // If you need to be compatible with both 1.5 and 1.7 versions then you can combine both writing
```

src/content/1.7/modules/core_updates/1.7.5.md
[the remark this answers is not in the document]
Quetzacoalt91: Of course!

jolelievre:
@Quetzacoalt91 good to go?

Quetzacoalt91 (approved):

PierreRambaud (approved):

PR #172, opened 2018-12-11
fixed changedCheckoutStep event name

adviva:
Checkout step change event name is wrong (see https://github.com/PrestaShop/PrestaShop/blob/develop/themes/_core/js/checkout.js#L63)


PierreRambaud:
Hey, Thanks @adviva  and nice catch :)

PR #185, opened 2019-01-14
Changing after PHP 5.4

sebastienhouzet:
Configuration not needed because hase removed since php 5.4

register_globals = Off
magic_quotes_gpc = Off


Quetzacoalt91:
Interesting, I see that there are no reference to the PDO extension, more suitable than the simple MySQL one.

Thank you @sebastienhouzet

PR #196, opened 2019-01-29
Fix file name in template-inheritance example

SharakPL:
The same file in the example is called with 3 different names: layout-pack.tpl, product-pack.tpl and product-page.tpl. Let's clear out confusion and stick with the one from img/template-inheritance.png so product-pack.tpl
+ minor improvement ;)


mickaelandrieu:
Agree, it's easier to understand this way: thanks @SharakPL!

PR #206, opened 2019-02-12
Fix Wrong method name

nenes25:
The method hookDisplayAdministrationPageForm should be named hookActionAdministrationPageForm like the real hook name 
Cf. "action{$this->hookName}Form" in src/Core/Form/FormHandler.php 
It doesn't works with hookDisplayAdministrationPageForm


mickaelandrieu:
Hello @nenes25,

you're right, thanks for the fix :+1: 

Mickaël

PR #226, opened 2019-03-06
Add deprecation notices

eternoendless:
Some hooks have been deprecated since 1.7.0. This PR adds deprecation notices to the ones that are documented.

See https://github.com/PrestaShop/PrestaShop/blob/develop/classes/Hook.php#L87-L104


mickaelandrieu:
Thank you @eternoendless

PR #228, opened 2019-03-11
Fixed Naming Conventions devdocs sections

lel-amri:
The meaning of "prefix" and "suffix" where swapped at some places


mickaelandrieu:
Thanks @lel-amri!

PR #232, opened 2019-03-18
Document CQRS

sarjon:
I referred to https://github.com/PrestaShop/PrestaShop/pull/9370 and found that explanation well written, so I basically ended up copying @eternoendless work with some minor changes to it. :smiling_imp: 

I have a few questions:

1. Should I provide code examples of how to create and use command/handler? Or maybe it should be subsection of CQRS?
2. Is `Architecture` good section for this article? I was also thinking about `Architecture/Migration Guide`, but it feels a bit wrong.


mickaelandrieu:
Nice introduction!

For me the docs should answer some questions of the PrestaShop developers:

* what is CQRS? (your introduction);
* am I concerned by this stuff when working on themes or modules? (tutorials, and/or workflows) Imagine you need to explain that in training for instance;
* what are the pros/cons of this choice for the PrestaShop developers (ie: why we decide to do that and how it helps PrestaShop developers to work with our platform?);

WDYT? /c  @eternoendless

matks:
> 1. Should I provide code examples of how to create and use command/handler? Or maybe it should be subsection of CQRS?

In another subsection, yes 😉it's 2 subtopics: what is X and how we implement X in PrestaShop. Like "what is Dependency Injection" and "How Symfony implements Dependency Injection with the Symfony Container"

> 2. Is `Architecture` good section for this article? I was also thinking about `Architecture/Migration Guide`, but it feels a bit wrong.

Looks like the right place to me 👍

src/content/1.7/development/architecture/cqrs.md
```diff
@@ -0,0 +1,69 @@
+---
+title: CQRS
+weight: 10
+---
+
+# CQRS
+
+## What is CQRS?
+
+CQRS stands for _Command Query Responsibility Segregation_. In brief, the CQRS pattern suggests to separate "write" and "read" models, which it refers to as Commands and Queries.
```
matks:
```suggestion
CQRS stands for _Command Query Responsibility Segregation_. In brief, the CQRS pattern suggests to separate "write" and "read" models, which it refers to as Commands and Queries. In a web application like a CMS, we perform either "read" operations which return information to the user or "write" operations which modify the data managed by the application.
```

src/content/1.7/development/architecture/cqrs.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Why using CQRS in PrestaShop?
+
```
matks: Can you explain why we need to keep core and why we need to isolate it ?
Something like

"We are currently migrating PrestaShop Back Office to Symfony but the new Controllers, Templates and classes dedicated to the BO are still using the Core classes of PrestaShop used by the previous BO. One day, we will remove this Core too. Until that day, we need to allo the new BO to access these classes and data while keeping a clear border between these 2 areas of code as one is expected to remain and the other one to disappear."

src/content/1.7/development/architecture/cqrs.md
```diff
@@ -0,0 +1,69 @@
...
+public function handle(NameOfTheQuery $query);
+```
+
+## Command and Query buses
+
+Command bus is a pattern used to map `Commands` and `Queries` to `CommandHandlers` and `QueryHandlers`. PrestaShop defines it's own `CommandBusInterface` and implements it using [Tactician command bus library](https://tactician.thephpleague.com/).
+
+PrestaShop uses 2 commands buses:
+
+1. `CommandBus` - for dispatching `Commnads` only
```
matks:
```suggestion
1. `CommandBus` - for dispatching `Commands` only
```

src/content/1.7/development/architecture/cqrs.md
```diff
@@ -0,0 +1,69 @@
...
+CQRS stands for _Command Query Responsibility Segregation_. In brief, the CQRS pattern suggests to separate "write" and "read" models, which it refers to as Commands and Queries.
+
+## Why using CQRS in PrestaShop?
+
+During Back Office migration to Symfony, PrestaShop needs a way to access and alter data on the new Symfony pages without multiplying the sources of truth, and without accessing `ObjectModel` (e.g. `Product`, `Category`) directly.
+
+Commands and Queries allow us to isolate the controllers from the data source, which can be later replaced by something else while leaving behind a nice API.
+
+This implementation proposes a "top-down" design, which inverses the classic data-driven design. It starts on a page and the actions performed in it, and then trickles down layer by layer, finishing on the data layer.
+
```
matks: It could be nice to have a schema here. This schema would compare old and new architecture.

In old architecture we would see the browser sending HTTP requests to a big bunch of code which would contain Controllers calling directly core classes. Like:

browser ==(send HTTP requests)==> controller ==(calls)==> core

In new architecture, we would show the clear separation of the bus

browser ==(send HTTP requests)==> SF controller =(dispatchs Queries and Commands)=> Bus routing =(Handler handles the Command/Query)=> Adapter Handler ==(calls)==> core

src/content/1.7/development/architecture/cqrs.md
sarjon: Do you think this commit https://github.com/PrestaShop/docs/pull/232/commits/b908ef4eff4effd4a9170531bd0ebe171f3f5c5c explains it?

src/content/1.7/development/architecture/cqrs.md
sarjon: Interesting, I've tried to do so in https://github.com/PrestaShop/docs/pull/232/commits/b908ef4eff4effd4a9170531bd0ebe171f3f5c5c commit, could you check?

matks: 👍

matks:
Thank you @sarjon

PR #239, opened 2019-04-01
Add deprecation notice about PS 1.7.6

Quetzacoalt91:
This new page deals with the new CLDR introduced on PS 1.7.6 and warns about the removed stuff from the old one.


mickaelandrieu:
Thanks @Quetzacoalt91 !

PR #242, opened 2019-04-10
Add "Insomnia" to test REST API

yannicka:
As it is a FLOSS (and good) software, I think we can add "Insomnia" to test REST API


matks:
Thank you @yannicka

PR #244, opened 2019-04-10
Documents Bulk actions for Grid component

sarjon:


src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
```diff
@@ -1,6 +1,142 @@
...
 
-> Do you know how to do it? Please feel free to contribute to the docs!
\ No newline at end of file
+# How to use Bulk actions in Grid?
+
+In addition to displaying data tables, Grid component also comes with a lot of additional functionality. One of it is Bulk actions.
+
+## What is Bulk action?
+
+Bulk action is action (e.g. Delete, Change status & etc) that can be performed on multiple entries in table. This is a common task when using Grid (e.g. Change status for multiple Orders in single action).
```
matks:
```suggestion
Bulk action is action (e.g. Delete, Change status & etc) that can be performed on multiple entries in table. This is a common task when using Grid (e.g. Delete multiple selected Orders with a single click).
```
I think this example is really widely known and used 🤔

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
```diff
@@ -1,6 +1,142 @@
...
+                ->setOptions([
+                    'customers_bulk_delete_route' => 'admin_customers_delete_bulk',
+                ])
+            )
+        ;
+    }
+}
+```
+
+As you see, Grid stores Bulk actions in `BulkActionCollection` object. This mean that every Bulk action needed for Grid must be added to `BulkActionCollection` using `add()` method.
```
matks:
```suggestion
As you see, Grid stores Bulk actions in `BulkActionCollection` object. This means that every Bulk action needed for Grid must be added to `BulkActionCollection` using `add()` method.
```

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
```diff
@@ -1,6 +1,142 @@
...
+
+## How to add Bulk actions from module?
+
+You can add additional Bulk actions to any Grid using hook. You have to follow these steps:
```
matks: 👍

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
```diff
@@ -1,6 +1,142 @@
...
+        ;
+    }
+}
+```
+
+Now you should be able to see new Bulk action available in Customers grid!
+
+## How to create custom Bulk action?
+
+PrestaShop already comes with a bunch of Bulk actions that are available for use in your Grids or for extending PrestaShop grids. However, in some use cases you may find that exisitng Bulk actions does not fit your needs. Luckly, there's a solution, you can create custom Bulk actions.
```
matks:
```suggestion
PrestaShop already comes with a bunch of Bulk actions that are available for use in your Grids or for extending PrestaShop grids. However, in some use cases you may find that existing Bulk actions do not fit your needs. Luckily, there is a solution, you can create custom Bulk actions.
```
Please avoid "let's" or "there's" 😉I think this might not be suitable for a doc which aims to be accurate and exhaustive rather than friendly

Not sure about this though @LouiseBonnard wdyt ?

matks:
👏 

Do you think it would be relevant here to quickly explain how BulkActions require a dedicated Controller to handle the route they link to (for example SubmitBulkAction has a 'submit_route' option that needs to be given a valid SF route name to handle the submit process)

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
LouiseBonnard: Absolutely, please avoid abbreviated forms when writing an official text. ;-)

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
mickaelandrieu: agree, commited :)

sarjon:
I think that should be part of another section. As this is required for other actions as well.

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
sarjon: I can't accept your change as `Delete` action does not exist for Order in PrestaShop. :smiling_imp:

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
sarjon: Ok, but I can find these words in other documentations, like Symfony.

src/content/1.7/development/components/grid/tutorials/work-with-bulk-actions.md
matks: Aaaargh indeed 😅. "Delete bulk suppliers" then ? I thought that people often need to delete things in batches (like after having creating multiple dummies for testing purpose)

sarjon: Done with suppliers.

sarjon:
@matks addressed all your suggestions.

matks:
Thank you @sarjon

PR #248, opened 2019-04-11
Add deprecation notice

dennispw:
Since it's not maintained any more and haven't been for quite some time.


PierreRambaud:
Could be a good idea to find some people from the community who want to maintain this repository.
@Matt75 @PrestaShop/prestashop-core-developers wdyt guys? 

Yes @Matt75 , I add you because you can give us a lot of feedback / idea :smile:

mickaelandrieu:
For now, the statement is true.

So, until we find some community support, I think it's ok to merge it.

Thanks for the contribution @dennispw!

PR #251, opened 2019-04-11
Update _index.md

budhead:
Mention that developers should run composer dumpautoload after they put their own composer file in their module.


mickaelandrieu:
Thanks @budhead !

PR #255, opened 2019-04-16
Document domain exceptions

rokaszygmantas:


src/content/1.7/development/architecture/domain-exceptions.md
```diff
@@ -0,0 +1,138 @@
+---
+title: Domain exceptions
+weight: 10
+---
+
+# Domain exceptions
+
+With introduction of domain driven design in PrestaShop, a new way of error handling in the code is seeing the daylight.
+The code in `PrestaShop\PrestaShop\Core\Domain` namespace started to utilize *domain exceptions*.
```
matks: 🌈

src/content/1.7/development/architecture/domain-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // ...
+}
+```
+
+So now, if we are editing a category and it fails for some reason (e.g. the database is not responding),
+the controller will catch the `CannotEditCategoryException` (which is a child of `CategoryException`) and display a specific error message to the user.
+
+## But there are so many new exception classes. I'm scared!
```
matks:
```suggestion
## Why are there so many new exception classes.
```
Maybe a bit more professional 😛

rokaszygmantas: oh I thought I reached the pinnacle of poetry :sweat_smile:

src/content/1.7/development/architecture/domain-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+    // ...
+}
+```
+
+So now, if we are editing a category and it fails for some reason (e.g. the database is not responding),
+the controller will catch the `CannotEditCategoryException` (which is a child of `CategoryException`) and display a specific error message to the user.
+
+## Why are there so many new exception classes.
+
+**Don't be!**
```
matks:
```suggestion
```

matks:
Thank you @rokaszygmantas

PR #256, opened 2019-04-17
Adds documentation about value objects

sarjon:


sarjon:
@matks I've put it under `Core development > Domain > Value Objects`. However, there is `Core development > CQRS`, but I feel like `Domain` is more appropriate section for this docs. 

Now I'm thinking where DTOs should be placed? Maybe `CQRS` section should be moved under `Core development > Domain`? 

This structuring is crazy. :boom:

src/content/1.7/development/architecture/domain/_index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Domain
+menuTitle: Domain
+weight: 20
+---
+
+# Domain
```
mickaelandrieu: this has no sense to commit, then publish, then provide to Google and people empty pages, don't you think?

sarjon: Yeah, see my question about structuring above, I'm not sure where it should belong. When I get feedback, I'll update/change it. :)

eternoendless:
There already is a [Core Development > Architecture > Domain](https://devdocs.prestashop.com/1.7/development/architecture/domain/) section, you can put it under that

eternoendless:
ping @sarjon, could you please rebase?

sarjon:
@eternoendless done!

src/content/1.7/development/architecture/domain/value_objects.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Value Objects
+weight: 20
+---
+
+# Value Objects
+
+PrestaShop is using Value Objects in its codebase, to help you understand better what they are, this article aims to explain what Value object is and how to use it.
```
eternoendless:
```suggestion
PrestaShop is using Value Objects in its codebase. To help you understand better what they are, this article aims to explain what Value object is and how to use it.
```

src/content/1.7/development/architecture/domain/value_objects.md
```diff
@@ -0,0 +1,59 @@
...
+
+## What is Value Object?
+
+A Value Object is small immutable object whos equality is not based on identity, this means that two Value Objects are equal when they have the same value, but not necessarily being the same object.
```
eternoendless:
```suggestion
[Value Objects](https://martinfowler.com/bliki/ValueObject.html) (VO for short) are small, simple, immutable objects, like Money or a date range, whose equality is not based on their identity. This means that any two given instances of a same VO are considered equal when they contain the same values.

Value Objects are often referred to as Value Types as well. You can consider a Value Object like a data _type_, whose objective is to provide structure, validation, and meaning.
```

src/content/1.7/development/architecture/domain/value_objects.md
```diff
@@ -0,0 +1,59 @@
...
+
+Characteristics of Value Object are:
+
+* It does not have identity (e.g. `Money` is value object because it can be created without id like `new Money($amount)`).
```
eternoendless:
```suggestion
* It does not have an identity, contrary to an Entity (e.g. `Money($amount)` can be considered a VO because two instances of it can be considered equal as long as the `$amount` is equal, regardless of which bills that amount is made of; but `Bill($billNumber)` is an entity because it refers to a specific physical Bill).
```

src/content/1.7/development/architecture/domain/value_objects.md
```diff
@@ -0,0 +1,59 @@
...
+* It is immutable (e.g. it cannot be modified, but instead new instance of value object is created after its modification).
```
eternoendless:
```suggestion
* It is immutable (e.g. it cannot be modified; any modification on it yields a different instance).
```

src/content/1.7/development/architecture/domain/value_objects.md
```diff
@@ -0,0 +1,59 @@
...
+
+As an example, we can take a look at `PrestaShop\PrestaShop\Core\Domain\Currency\ValueObject\ExchangeRate` value object. `ExchangeRate` is used to pass currency exchange rate between different parts of system.
+
+```php
+use PrestaShop\PrestaShop\Core\Domain\Currency\ValueObject\ExchangeRate;
+
+// ExchangeRate has VO has rules to protect.
+// In this case, ExchangeRate validates that it's value cannot be 0 or less.
+// If we were to create ExchangeRate with value of -1,
+// then CurrencyConstraintException would be thrown.
```
eternoendless: I think it's better to leave long texts outside the code section

sarjon: Ummm, I see that Symfony docs are using comments to explain code.

In addition, I believe that many developers just skip reading documentation, but dive straight into code example, so even if it duplicates content outside the code, it's not necessarily a bad thing imo.

But I can remove it if you wish. :)

PierreRambaud: Agreed with @sarjon :roll_eyes:

matks:
Thank you @sarjon

PR #257, opened 2019-04-18
Document DTOs

sarjon:


sarjon:
@matks ready for review.

src/content/1.7/development/architecture/domain/data-transfer-objects.md
```diff
@@ -0,0 +1,52 @@
...
+     * Handler returns EditableCustomer DTO instead of unstructured array,
+     * thus allowing developer to know exactly what data can be accessed.
+     *
+     * @return EditableCustomer
+     */
+    public function handle(GetCustomerForEditing $query);
+}
+```
+
+Now image you are using implementation of `GetCustomerForEditingHandlerInterface` handler in your code.
```
matks:
```suggestion
Now imagine you are using implementation of `GetCustomerForEditingHandlerInterface` handler in your code.
```

src/content/1.7/development/architecture/domain/data-transfer-objects.md
```diff
@@ -0,0 +1,52 @@
...
+weight: 20
+---
+
+# Data Transfer Objects
+
+PrestaShop is using Data Transfer Objects (DTO in short) in its codebase, in this section you learn how you can benefit from using DTOs as well.
+
+## What is Data Transfer Object?
+
+Data Transfer Object is object (e.g. `UserData`) that contains data in a structured format. This means that DTO usually comes with getter methods or public properties for accessing it's data. The main benefit of using DTOs is that you know what data it has.
```
matks: You can also add that DTOs:
- enables type hinting
- are valid because built (you cannot "skip" one of the keys)
- provide better performance on data access than arrays since php5.5 (at least)
- use less memory (useful for big data collections)
See https://www.brandonsavage.net/stop-returning-arrays-use-objects-instead/ for details 😉

sarjon: > provide better performance on data access than arrays since php5.5 (at least)
> use less memory (useful for big data collections)

Regarding these 2, I don't think we should state that, it may really depend on how you use DTOs, maybe you have array of DTOs inside DTO. So I'd suggest skipping them. 

Regarding others, sure, we could add them to docs as well. :)

matks:
Thank you @sarjon

PR #259, opened 2019-04-24
Improve PHP compatibility chart

eternoendless:
Improve PHP compatibility chart. It requires HTML but it looks better :)

Screenshot 2019-04-24 at 13 48 19


matks:
Thank you @eternoendless

PR #262, opened 2019-04-30
All calls to this hook don't pass the $_POST object.

manfield:
Although this would be very useful it is not like this in this moment.


mickaelandrieu:
Hi @manfield,

would you consider to provide a patch in the core to make the docs true instead? 🙏 

Regards,

PierreRambaud:
@mickaelandrieu We don't use `_POST` index in the core, and the $_POST is already a global variable.
So the fix must be done in the doc :)

mickaelandrieu:
Thanks for the contribution @manfield !

PR #266, opened 2019-05-04
Core style theme.css priority should be 50 not 0

rateforx:
https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/FrontController.php#L906


mickaelandrieu:
Hello,

you're right!

Thanks 👍

PR #269, opened 2019-05-17
Email theme generation documentation

### LINKED ISSUE #208: Add new hooks from EmailGenerator
Add new hooks from EmailGenerator
see https://github.com/PrestaShop/PrestaShop/pull/12046

jolelievre:
Introduce documentation about the "Email theme" component

- general description of the feature
- detailed workflow
- available hooks
- tutorials on how to use the feature in your modules
- example module

Fixes https://github.com/PrestaShop/docs/issues/208


src/content/1.7/development/components/email-theme/_index.md
```diff
@@ -0,0 +1,183 @@
+---
+title: The Email Theme component
+menuTitle: Email Theme
+weight: 2
+---
+
+# The Email Theme component
```
sarjon: Wouldn't it make more sense to call it `Mail Template component` to follow same naming that is in code?

jolelievre: This would we logical from a developper perspective, but the page in the back office is named "Email theme" so I thought matching with this would be more "user friendly"

jolelievre: Besides the word "template" is a bit confusing in this situation, I had to add a section `Vocabulary` in this doc to avoid confusion between layout and template.. not really easy to find correct and clear naming on this feature

sarjon: Well, it's developer documentation, so calling same component in different names is weird imo. It's the same as calling `Grid component` in code, but `List component` in docs.

eternoendless:
Please consider moving the tutorials to the Modules section

jolelievre:
> Please consider moving the tutorials to the Modules section

That's fine for me, but I don't know in which subpart I could put it. The architecture of the modules section isn't very clear. And although this doc contains some examples about extending the themes via a module there are also more generic info about the generator. Which will be probably extended when the feature is improved (configurable theme for example, along with how you set their configuration).

jolelievre:
@eternoendless Alright I moved the tutorials in the module section, the workflow and feature is still in the Components section, this seemed more logical to me.
And the example module has been merged as well https://github.com/PrestaShop/example_module_mailtheme
So if it's ok for you we can merge this doc as well!

PR #274, opened 2019-05-23
Explain how to override a simple block instead of a whole file

PierreRambaud:


src/content/1.7/modules/concepts/templating/admin-views.md
```diff
@@ -32,10 +32,13 @@ In a module called `foo` let's create the related template. As the template is l
...
 
+Note that we use `PrestaShopBundle` instead of `@PrestaShop` to be sure we extend the original file.
+
 ```twig
-{% extends '@PrestaShop/Admin/layout.html.twig' %}
-{% block content %}
-    Hello world!
+{% extends 'PrestaShopBundle:Admin/Product/CatalogPage:catalog.html.twig' %}
+
+{% block product_catalog_filters %}
```
matks: The statement later in the file (l.47) says "we have overridden the complete page" which is now false.

Should'nt we rather add a statement at the end of this page, saying "by the way, you can override only a twig block !" 😉

PierreRambaud: I improve next line. It's really better to play with block instead whole file.

matks:
Thanks @PierreRambaud

PR #280, opened 2019-05-29
Fixed hook name

mickaelandrieu:


matks:
Should'nt we fix it first in hook.xml and SQL upgrade file ?

mickaelandrieu:
I can't because docs and core are not in the same repositories ;)

Am I repeating myself? 👼

matks:
> I can't because docs and core are not in the same repositories ;)
> 
> Am I repeating myself? 👼

I dont see how this prevents us from doing a PR in the Core repository fixing the hook name in both the hook.xml file and the dedicated SQL upgrade first.

I can show you if you want :trollface:

mickaelandrieu:
> I dont see how this prevents us from doing a PR in the Core repository fixing the hook name in both the hook.xml file and the dedicated SQL upgrade first.

Think about it: if the docs were inside the core repository, I could address it in one step. Even if we do a PR right now, it will be merged in 1.7.7, so in almost 1 year for people :p

matks:
> Even if we do a PR right now, it will be merged in 1.7.7, so in almost 1 year for people :p

The fix needs only to target the hook.xml file (not critical) and the SQL upgrade (critical, but no code)
Considering this avoids some confusion about the hooks, and that could justify a fix in 1.7.6.x

ping @eternoendless

mickaelandrieu:
LGTM for the docs part ?

eternoendless:
Thank you @mickaelandrieu

PR #288, opened 2019-06-15
Improve doc for module's services.yml file

aziule:
I found the current doc to be misleading regarding how to handle a module's `config/services.yml` and thought it would be nice to explain to developers how things work (at least at a high level).


Matt75:
Have to document container for legacy Admin & Front too.

See example : https://github.com/PrestaShop/productcomments/tree/master/config

Please note legacy container doesn’t have Symfony services.

PR #305, opened 2019-07-15
Introduces number and integer min max filter type documentation

tomas862:


src/content/1.7/development/components/form/types-reference/integer-min-max-filter.md
```diff
@@ -0,0 +1,29 @@
+---
+title: IntegerMinMaxFilterType
+menuTitle: IntegerMinMaxFilterType
```
eternoendless: This is not needed if it's the same as the title

src/content/1.7/development/components/form/types-reference/integer-min-max-filter.md
```diff
@@ -0,0 +1,29 @@
...
+weight: 2
```
eternoendless: Weight is used for ordering. If all elements have the same weight then it's useless :)

tomas862: Thank you - had to do it for all form type-reference level .md files to keep them all aligned alphabetically . https://github.com/PrestaShop/docs/pull/305/commits/8d18a6047d4e7fe995a4ece4835e42ebfb35d423

eternoendless:
Thank you @tomas862

PR #310, opened 2019-07-18
Fix chmod 000

TemKaSD:
Fix chmod 000 in files appParameters.php and class_index.php.
https://github.com/PrestaShop/docs/issues/308

### DIFF
--- a/src/content/1.7/scale/optimizations.md
+++ b/src/content/1.7/scale/optimizations.md
@@ -101,6 +101,8 @@ opcache.enable_file_override = On
 opcache.interned_strings_buffer=64
 opcache.memory_consumption=256
 opcache.max_accelerated_files=20000
+opcache.validate_timestamps=1
+opcache.revalidate_freq=0
 ```
 
 Also, your favorite ecommerce project made sure it's fully compatible with OPCache.


src/content/1.7/scale/optimizations.md
```diff
@@ -101,6 +101,8 @@ opcache.enable_file_override = On
 opcache.interned_strings_buffer=64
 opcache.memory_consumption=256
 opcache.max_accelerated_files=20000
+opcache.validate_timestamps=1
```
PierreRambaud: This is the default value so not needed.

src/content/1.7/scale/optimizations.md
```diff
@@ -101,6 +101,8 @@ opcache.enable_file_override = On
...
+opcache.revalidate_freq=0
```
PierreRambaud: Not sure it's a good idea to revalidate cache on each request :thinking:

TemKaSD: > Not sure it's a good idea to revalidate cache on each request 🤔

Without these changes error 500, check.

TemKaSD: @PierreRambaud https://youtu.be/5rdthNnruFE

PierreRambaud: I'm not able to reproduce with nginx / apache2.
Are you sure your php have the same user as your web server? Which user / group have your ftp account?

PierreRambaud: If these configuration lines work for you, you can keep it, and if you have a problem we can't reproduce, we're not able to approve changes. 
But, if you can describe a little more in a troubleshooting section what you had, and how you fixed it (by adding these two lines), we'll be glad to approve it :)

TemKaSD: > I'm not able to reproduce with nginx / apache2.
> Are you sure your php have the same user as your web server? Which user / group have your ftp account?

Yes.
phpfpm:

user = mbm
group = mbm
listen.owner = mbm
listen.group = www-data

files:

user = mbm
group = mbm

TemKaSD: > If these configuration lines work for you, you can keep it, and if you have a problem we can't reproduce, we're not able to approve changes.
> But, if you can describe a little more in a troubleshooting section what you had, and how you fixed it (by adding these two lines), we'll be glad to approve it :)

Link pls troubleshooting section.

PierreRambaud: You need to create one :)

TemKaSD: > You need to create one :)

?

PierreRambaud: There is no Troubleshooting section here, you need to create a new one. Can be at the bottom of this file.

PR #315, opened 2019-07-30
add server parameter to registerjavascript

zalexki:
if someone has ideas about the wording of the description


Quetzacoalt91:
Reviewed with https://github.com/PrestaShop/PrestaShop/blob/7228af62726336bf25d544f95e2468d671121b29/classes/controller/FrontController.php#L1089

and https://github.com/PrestaShop/PrestaShop/blob/7228af62726336bf25d544f95e2468d671121b29/classes/assets/JavascriptManager.php#L47-L56

PR #321, opened 2019-08-08
Improve documentation on module translation

### LINKED ISSUE #284: Document translation system and legacy fallback

eternoendless:
Fixes #284


src/content/1.7/modules/creation/module-translation/classic-system.md
```diff
@@ -0,0 +1,377 @@
...
+  {assign var='my_module_name' value=$smarty.capture.my_module_tempvar}
+{/if}
+{l s='Hello,' mod='mymodule'} {$my_module_name$}!
+```
+
+This is a start! In this example, when `$my_module_name` is not defined, we capture the value of translating "World" and assign that value in `$my_module_name`. This way, we can simply display `$my_module_name` in any case.
+
+Now, when working in internationalization, it's better not to make assumptions regarding spacing and positioning of variables. For example, in French, there should be a non-breaking space before the exclamation mark (`!`), but not in other languages. How can we handle this?
+
+In PrestaShop, you can interpolate variables within your wordings using using [sprintf replacement markers](https://php.net/sprintf), such as `%s` or `%1$s`. Using this feature, we could get rid of the explicit concatenation of "Hello, ", "$my_module_name" and "!", and let it depend on the translated wording:
```
mickaelandrieu:
```diff
- using using
+ using
```

src/content/1.7/modules/creation/module-translation/classic-system.md
```diff
@@ -0,0 +1,377 @@
...
+
+Instead, you should use placeholders and replace them with HTML code.
+
+Let's take an example with a link in a string, which can be tricky to do. 
+The first solution that comes to mind would be to concatenate the translated strings with raw HTML code.
+But as we explained before, this solution is not recommended, because the words order could be different depending on the translation language.
+
+```php
+{l
+  s='If you want a category to appear in the menu of your shop, go to [1]Modules > Modules & Services > Installed modules.[/1] Then, configure your menu module.'
```
mickaelandrieu: couldn't we introduce a new best practice here? WDYT about that:

```
[modules-link]Modules > Modules & Services > Installed modules.[/modules-link]
```

I'm not convinced it will work as I hope.

src/content/1.7/modules/creation/module-translation/classic-system.md
```diff
@@ -0,0 +1,377 @@
...
+
+global $_MODULE;
+$_MODULE = array();
+$_MODULE['<{mymodule}prestashop>mymodule_2ddddc2a736e4128ce1cdfd22b041e7f'] = 'Mon module';
+$_MODULE['<{mymodule}prestashop>mymodule_d6968577f69f08c93c209bd8b6b3d4d5'] = 'Description du module.';
+$_MODULE['<{mymodule}prestashop>mymodule_533937acf0e84c92e787614bbb16a7a0'] = 'Êtes-vous certain de vouloir désinstaller ce module ? Vous perdrez tous vos réglages !';
+$_MODULE['<{mymodule}prestashop>mymodule_0f40e8817b005044250943f57a21c5e7'] = 'Aucun nom fourni';
+```
+
+This file shouldn't be modified manually! It is meant to be edited through the PrestaShop translation interface.
```
mickaelandrieu: you may put this advice inside a note or a warning block.

src/content/1.7/modules/creation/module-translation/new-system.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+Since the module is called MyModule, the translation domain should be `Modules.Mymodule.Mymodule`. The third part matches the file name, which is also "mymodule".
+
+#### Module controllers
+
+`ModuleAdminController` and `ModuleFrontController` can access the module instance via the `$this->module` property.
+
+```php
+// file: controllers/front/something.php
```
mickaelandrieu: question: what will happens if I have another file named something.php in my module?

mickaelandrieu (commented):
Awesome !

Can we also provide a module which is using the new system and still retro compatible with 1.6?

There is no need to be a big module: only a hello-world like that we could comment and share here.

src/content/1.7/modules/creation/module-translation/classic-system.md
eternoendless: Yes it should work 👍

src/content/1.7/modules/creation/module-translation/new-system.md
eternoendless: AFAIK they will both treated as if they were the same file

eternoendless:
> Can we also provide a module which is using the new system and still retro compatible with 1.6?

It's _possible_, but it won't be nice 😅. You'd need two copies of templates that use `{l}` smarty tag (one for 1.6 with `mod=` and one for 1.7.6+ with `d=`) and use one or the other according to the current PS version. You'd also have to either use different controllers or different implementations that use the `l()` function or the `trans()` function.

The new translation system won't pick up wordings used through `l()` either.

matks (approved):

matks:
Thank you @eternoendless

PR #327, opened 2019-08-22
Fix typo for 'modules' in 1.7/themes/getting-started/theme-organization

hdasdoria:
There's just a typo that could cause small problems when recreating themes folder structure for example.

In the docs, the folder was called 'module' instead of 'modules' in the directory structure explanation.


PierreRambaud:
Nice catch! Thanks.

PR #334, opened 2019-09-04
Explain how to use Symfony service overrides and decorations

matks:


jf-viguier:
Thanks a lot for that usefull doc @matks

kpodemski:
Do we have any section informing about current PrestaShop services? :) It would be nice to have it without having to look into the code.

matks:
> Do we have any section informing about current PrestaShop services? :) It would be nice to have it without having to look into the code.

I'm afraid there are hundred of them 😐 so it would be a veryyyyy long list

I think you ask for this because you have a need 😉 can you tell me what is this need ? Maybe there's another more efficient solution than a list for this need

PierreRambaud:
Guys, i'm stupid
@kpodemski You can use
```bash
./bin/console debug:container
```
To list all available services.

matks:
> Guys, i'm stupid

I dont know why you say this 😄 

But even the services list is not helpful. Maybe @kpodemski needs something more than a list ?

kpodemski:
@matks 

yeah i was asking for a PrestaShop specific services, sorry for not being precise ;-)

matks:
> @matks
> 
> yeah i was asking for a PrestaShop specific services, sorry for not being precise ;-)

There are still dozen of them 😅 we follow the [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) to make the code as simple and decoupled as possible (and avoid scary god classes).

The cool thing about it is that it's easy to customize using services override/decoration to customize **only the part you want** (instead of overriding the whole Product.php). The bad thing about it is it means a lot of services.

kpodemski:
@matks 

IMO the more the better in this case, thanks to SRP it's a way better method to override some parts of the core, the only downside is that we need to wait for a next major version to have proper services on the front :)

PierreRambaud:
@kpodemski We are waiting for this too :sob: With twig in FO, could be awesome.

src/content/1.7/modules/concepts/services/_index.md
```diff
@@ -154,6 +161,82 @@ If you need more details about dependency injection and how services work in the
...
+To do it: you declare your new service using the old service name. So if you want to override the service `prestashop.core.b2b.b2b_feature` with your own implementation, you write in `config/services.yml` :
+
+```yml
+  prestashop.core.b2b.b2b_feature:
+    class: 'YourCompany\YourModule\YourService'
+```
+
+That's done. The service registered under the name `prestashop.core.b2b.b2b_feature` is now your service. The previous `prestashop.core.b2b.b2b_feature` is gone.
+
+#### Decorate the service
```
jolelievre: I have a question, about decoration Is it possible for multiple modules to decorate the same service
Will the decoration stack on top of each other then?
If so then we should encourage using this method as it would allow multiple modules to coexist.

The problem we have today with class override will be the same with service override, so anyway we should add a warning notice about this drawback no?

src/content/1.7/modules/concepts/services/_index.md
```diff
@@ -154,6 +161,82 @@ If you need more details about dependency injection and how services work in the
...
+
+To do it: you declare your new service using the 'decorates' keyword. So if you want to decorates the service `prestashop.core.b2b.b2b_feature` with my own implementation, you write in `config/services.yml` :
+
+```yml
+ my_own_b2b_feature_service:
+    class: 'YourCompany\YourModule\YourService'
+    decorates: 'prestashop.core.b2b.b2b_feature'
+```
+
+That's done. The service registered under the name `prestashop.core.b2b.b2b_feature` is now your service. The previous `prestashop.core.b2b.b2b_feature` is still available under the name `prestashop.core.b2b.b2b_feature.inner`
```
jolelievre: Is the `inner` suffix an automatic feature from Symfony?

src/content/1.7/modules/concepts/services/_index.md
```diff
@@ -154,6 +161,82 @@ If you need more details about dependency injection and how services work in the
...
+
+The decoration strategy can be very useful if:
+
+- you want some areas of your code to use the new service, some others to use the old service
+- you want to use the old service in the implementation of the new service
```
jolelievre: But how do you force using the old service? With the inner name?
Then we agree that in all other places of the code, the new service will be used without the codebase being aware of it?

src/content/1.7/modules/concepts/services/_index.md
```diff
@@ -154,6 +161,82 @@ If you need more details about dependency injection and how services work in the
 their documentation about the [Service Container](https://symfony.com/doc/3.4/service_container.html).
 {{% /notice %}}
 
+### Override an existing Symfony service
+
+The container definition can be modified by a module, which enables you to override an existing Symfony service being used in PrestaShop.
+
+This is a mechanism similar to PrestaShop standard overrides, but the main benefit is that the php code stays unmodified. This prevents issues linked to code definition or autoloading failures.
+
+As you can read it from the [Symfony documentation](https://symfony.com/doc/current/service_container/service_decoration.html), there are 2 ways to modify an existing service:
```
jolelievre: Maybe you could add a paragraph about interfaces, explaining that we rely more and more on them and that overridden services should implement the appropriate service rather than extending an existing class.
Although sometimes you won't have a choice but to extend it (because of type checking).

src/content/1.7/modules/concepts/services/_index.md
matks: Yes

src/content/1.7/modules/concepts/services/_index.md
matks: Yes indeed, this is the Symfony behavior
So if you want to still use the old service, you need to modify the places where you want to use it and inject `[my-service].inner` instead

src/content/1.7/modules/concepts/services/_index.md
matks: Indeed, whether it's PrestaShop class override or Symfony service override, the mechanism is powerful and consequently can break a shop.

Yes, I think you can decorate multiple times a service 🤔 but I dont know which one will get aliased to `inner` 😅

src/content/1.7/modules/concepts/services/_index.md
matks: 👍

src/content/1.7/modules/concepts/services/_index.md
matks: In fact, service decoration or service override are both better than PrestaShop class override because **the php class remains untouched**.

So if we have 2 services which use the same class:
```
    my_service_one:
        class: 'MyClass'

    my_service_two:
        class: 'MyClass'
```
Then overriding or decorating `my_service_one` will not break or change the behavior of `my_service_two`. PrestaShop standard override mechanism will do it, on the contrary.

PR #340, opened 2019-09-11
Fix a spelling mistake in the doc

senkuu:


matks:
Thank you @senkuu ! nice catch 😉

PR #347, opened 2019-09-17
Explain how to comply with the conditions about using iFrames in a module

Quetzacoalt91:


src/content/1.7/modules/sell/techvalidation-keysteps.md
```diff
@@ -34,7 +34,10 @@ After meeting each requirement provided by the Validator, here's a list of extra
 * The presence of DROP/ALTER of PrestaShop tables.
     * It's highly forbidden to apply any changes on PrestaShop Core tables. It's very very dangerous and we don't allow any risk.
 
-* Except for [Payment Modules](https://github.com/PrestaShop/paymentexample/blob/master/paymentexample.php#L150), the use of iframes is STRICTLY FORBIDDEN.
+* The use of iframes is highly discouraged for security reasons, although they are implemented in different part of the core like in [Payment Modules](https://github.com/PrestaShop/paymentexample/blob/master/paymentexample.php#L150).
+    * Using an iframe authorizes to load content from a site that is not controlled by us. This is the same problem as authorizing to load javascript file from an external source. If you're being hacked, the attacker could potentially exploit other failures to take control of all the shops that would have installed the module.
```
PierreRambaud:
```suggestion
    * Using an iframe authorizes to load content from a site that is not controlled by the PrestaShop app. This is the same problem as authorizing to load a javascript file from an external source. If the file is being hacked, the attacker could potentially exploit other failures to take control of your shop and all the shops that would have installed the module.
```

src/content/1.7/modules/sell/techvalidation-keysteps.md
```diff
@@ -34,7 +34,10 @@ After meeting each requirement provided by the Validator, here's a list of extra
...
+
+    * Therefore we need to check what your processes are, to ensure the security of the content that will be injected by this iframe into all the shops that will install the module. When submitting your module, the validation team will review the reasons why an iFrame is needed for this business and what are the measures taken by the provider to prevent attacks.
```
PierreRambaud:
```suggestion
    * Therefore we need to check what your processes are, to ensure the security of the content that will be injected by this iframe into all the shops that will install the module. When submitting your module, the validation team will review the reasons why an irame is needed for this business and what are the measures taken by the provider to prevent attacks.
```

src/content/1.7/modules/sell/techvalidation-keysteps.md
Quetzacoalt91: What about `If the source is being hacked` ?

Quetzacoalt91: As we talk to a module developer and not a merchant, I unfortunately disagree with the change `your shop`

PierreRambaud: GTM :+1:

Quetzacoalt91: Updated

matks:
Thank you @Quetzacoalt91

PR #348, opened 2019-09-18
Move composer content in a dedicated chapter

Quetzacoalt91:
This PR move the composer setup in a dedicated chapter in order to have a direct link to this feature.


matks:
Thank you @Quetzacoalt91

PR #350, opened 2019-09-20
Error in registerJavascript properties description

jeckyl:


matks:
Nice catch !

Thank you @jeckyl

PR #353, opened 2019-09-23
Better structure for module extension symfony concepts

matks:
Right now, clicking on "https://devdocs.prestashop.com/1.7/modules/concepts/" you are directly prompted Symfony extension concepts instead of a summary of all possible extension concepts. This PR fixes that.

### DIFF
--- a/src/content/1.7/modules/concepts/_index.md
+++ b/src/content/1.7/modules/concepts/_index.md
@@ -6,21 +6,4 @@ chapter: true
 
 # Extension concepts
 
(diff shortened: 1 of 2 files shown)


matks:
Thanks @Quetzacoalt91 for review !

matks:
It's better now 😊 next step will be to gather all Symfony extension concepts into a dedicated chapter. Right now they are splattered.

https://devdocs.prestashop.com/1.7/modules/concepts/symfony/

PR #354, opened 2019-09-24
Update admin controllers documentation

mrbig00:
Add `prepend-autoloader` to the example composer.json as mentioned at https://devdocs.prestashop.com/1.7/modules/concepts/composer/#important-notes


Quetzacoalt91:
Thank you @mrbig00

PR #361, opened 2019-10-01
Add notes that db account with full privileges have to be used

Mirgen:
Due to my personal experiences with upgrades on some web providers I find it nice to remind others that this can happen.


matks:
Interesting 🤔 thanks for the feedback ! So I guess the 2nd database account had only READ/WRITE access but could not create or update the SQL schema ?

Mirgen:
Exactly. It is obviously for security reasons. The bigggest web provider Wedos in Czech Republic have it like this.

matks:
Thank you @Mirgen

Mirgen:
@matks  BTW I am amazed how fast it was with your reply and merge 👍

matks:
@Mirgen Thanks 😄

PR #379, opened 2019-10-09
Reorganize contribute section, explain how to contribute with localhost

matks:
Following a suggestion from @jf-viguier 😉


jf-viguier:
Thanks @matks, it's important to better communicate on how simple is it to contribute.

PR #385, opened 2019-10-16
Improve hook example for form hooks usage

cesarmuriana:
In Adding new form field to customer form we need SwitchType, so I add this line "use PrestaShopBundle\Form\Admin\Type\SwitchType;"


matks:
Thank you @cesarmuriana

PR #390, opened 2019-10-28
Add link to system requirements tool

eternoendless:
Added information about PhpPsInfo which is very handy for not-that-technical people.


matks:
Thank you @eternoendless

PR #391, opened 2019-10-31
Update Hugo to 0.59.0

eternoendless:
Note that when running Hugo there is a warning related to this file: https://github.com/PrestaShop/docs/blob/master/src/themes/hugo-theme-learn/layouts/partials/menu.html#L144

It's a deprecation notice so nothing too serious for now.

Related theme issue here: https://github.com/matcornic/hugo-theme-learn/issues/329

### DIFF
--- a/bin/installHugo.sh
+++ b/bin/installHugo.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


matks:
@eternoendless do you need extensive tests ?

PierreRambaud:
Thanks @eternoendless

PR #395, opened 2019-11-04
Fix shortcodes following changes in Hugo 0.55

eternoendless:
Following [a change in Hugo 0.55](https://gohugo.io/news/0.55.0-relnotes/#shortcodes-revised), content in shortcodes (notice, funcdef, callout...) weren't being parsed to markdown anymore.

The solution, [as found here](https://discourse.gohugo.io/t/misunderstanding-of-shortcode-syntax-in-0-55/18538), is to pass the contents through the markdownify function.

I verified that it's working correctly.


PierreRambaud:
Thanks  @eternoendless

PR #399, opened 2019-11-07
Typo: set _PS_MODE_DEV_ to false instead of true

yannicka:


PierreRambaud:
Nice catch thanks @yannicka

Issue #398, opened 2019-11-06
Provide properties available in (Admin & Front) Module Controllers 


Quetzacoalt91:
Module developper can't guess the properties available in a ModuleAdminController or ModuleFrontController while developping.

For instance, we saw that a controller instanciated its own Module, although it was already available as `$this->module`

zalexki:
let's provide an example with `$this->module` and talk about the reflectionclass and `dump` to get more information ?

Quetzacoalt91:
An example with `$this->context` and `$this->module` would be great, these are the two main attributes used when developing a controller.

We can provide a generic way to get these attributes, as you suggest, but I wonder if all the details will be interesting for the developer. We may give it a try. :)

PR #402, opened 2019-11-08
Provide module and context properties in ModuleFrontController

Quetzacoalt91:
Fixes partially https://github.com/PrestaShop/docs/issues/398


zalexki:
you say ``$this->context` is a shortcut to the `\Context` class.` i think people will tend to not use it because it's `just a shortcut` maybe we can rephrase to incite people to use this instead of getContext.

Quetzacoalt91:
Indeed, `Context::getContext` has been specified

zalexki:
i meant saying something like `$this->context` is preferred to `Context::getContext()`.  has it avoid calling getContext if it's already set and we want them to use `$this->context` over `getContext()` AFAIK

PR #407, opened 2019-11-19
Update enabling-auto-update.md

idnovate:
According to `Addons Validator Team`, this is the *only* way to name upgrading scripts


PierreRambaud:
Thanks @idnovate

PR #408, opened 2019-11-20
Update _index.md

guirou62:
I made these changes to be coherant with these pages :
- https://devdocs.prestashop.com/1.7/modules/concepts/doctrine/
- https://github.com/friends-of-presta/demo-cqrs-hooks-usage-module/tree/master/src

Maybe, we can improve by adding Admin or Front


matks:
Thank you @guirou62

PR #416, opened 2019-12-02
twig file syntax error fixed

dheerajwebkul:


PierreRambaud:
Thanks @dheerajwebkul

PR #420, opened 2019-12-04
Update configuration.md

juguigui:
changes in Perfomances page / PSVersion without update docs

### DIFF
--- a/src/content/1.7/basics/installation/configuration.md
+++ b/src/content/1.7/basics/installation/configuration.md
@@ -22,14 +22,11 @@ Forcing the compilation of Smarty will always slow down the loading time of the 
 
 ## Displaying error messages
 
-PrestaShop’s default settings prevent the customer to see any server error message or any debugging code.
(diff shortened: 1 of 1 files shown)


src/content/1.7/basics/installation/configuration.md
```diff
@@ -22,14 +22,11 @@ Forcing the compilation of Smarty will always slow down the loading time of the
 
 ## Displaying error messages
 
-PrestaShop’s default settings prevent the customer to see any server error message or any debugging code.
+PrestaShop’s default settings prevent the customer to see any server error message or any debugging code. You, on the other hand, need this information in order to correct any potential mistake in your code.
```
matks: Hi ! I think your contribution is good but we should provide both ways to activate debug mode. Sometimes a user has broken its shop so bad that he cannot access the BO so the "php file" way can be useful too 😄

PierreRambaud:
Hi @juguigui ,

Since we had no news from you for more than 30 days, I'll close this pull request. Feel free to reopen or open another one if you think it's still relevant.

Thanks!

PR #421, opened 2019-12-06
Modules: replace "addCSS" and "addJS" by "registerStylesheet" and "registerJavascript"

yannicka:
`addCSS` and `addJS` are deprecated. Instead we should use `registerStylesheet` and `registerJavascript`.

### DIFF
--- a/src/content/1.7/modules/creation/displaying-content-in-front-office.md
+++ b/src/content/1.7/modules/creation/displaying-content-in-front-office.md
@@ -41,7 +41,7 @@ Attaching code to a hook requires a specific method for each:
 -   `hookDisplayRightColumn()`: will simply do the same as
     `hookDisplayLeftColumn()`, but for the right column.
 -   `hookDisplayHeader()`: will add a link to the module's CSS file,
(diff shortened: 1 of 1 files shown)


eternoendless:
Be aware that this works only for `ModuleFrontControllers`. `ModuleAdminControllers` still need `addCSS` and `addJS`.

eternoendless:
@Quetzacoalt91 can you confirm?

Quetzacoalt91:
Indeed, `registerStylesheet` & `registerJavascript` are defined in FrontController, preventing BO controller to access them.

matks:
> Be aware that this works only for `ModuleFrontControllers`. `ModuleAdminControllers` still need `addCSS` and `addJS`.

The modified doc is about front controller so this PR seems valid anyway ?

Progi1984:
Ping @yannicka

yannicka:
We're on a page named "Displaying content on the front office" and the class we create extends `ModuleFrontController`, so I think this PR is fine.

Progi1984:
Thanks @yannicka

PR #422, opened 2019-12-11
Add ps 1.7.7 min php ver and remove IE support

jf-viguier:
Ps should remove IE support, it's too old, too buggy


matks:
Thank you @jf-viguier

PR #424, opened 2019-12-13
Fix example controller name to be more coherent and usage of translator

matks:
Fixed example controller name to fit with naming convention (see https://devdocs.prestashop.com/1.7/modules/concepts/controllers/front-controllers/)

Thanks to @BlackKerio for the hint

### DIFF
--- a/src/content/1.7/modules/creation/module-translation/new-system.md
+++ b/src/content/1.7/modules/creation/module-translation/new-system.md
@@ -140,7 +140,7 @@ Since the module is called MyModule, the translation domain should be `Modules.M
 ```php
 // file: controllers/front/something.php
 
-class MyModuleSomethingFrontController extends ModuleFrontController
+class MySomethingModuleFrontController extends ModuleFrontController
 {
     public function initContent()
     {
--- a/src/content/1.7/modules/creation/module-translation/new-system.md
+++ b/src/content/1.7/modules/creation/module-translation/new-system.md
@@ -140,7 +140,7 @@ Since the module is called MyModule, the translation domain should be `Modules.M
 ```php
 // file: controllers/front/something.php
 
-class MySomethingModuleFrontController extends ModuleFrontController
+class MymoduleSomethingModuleFrontController extends ModuleFrontController
 {
     public function initContent()
     {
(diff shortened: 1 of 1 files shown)


src/content/1.7/modules/creation/module-translation/new-system.md
```diff
@@ -140,11 +140,11 @@ Since the module is called MyModule, the translation domain should be `Modules.M
 ```php
 // file: controllers/front/something.php
 
-class MyModuleSomethingFrontController extends ModuleFrontController
+class MySomethingModuleFrontController extends ModuleFrontController
 {
     public function initContent()
     {
-        $this->title = $this->module->trans('My module title', [], 'Modules.Mymodule.Something);
+        $this->title = $this->trans('My module title', [], 'Modules.Mymodule.Something);
```
matks: According to my test, this is the right call. Can someone confirm, maybe @Quetzacoalt91 ?

PierreRambaud: I confirm you this is the right call (according to my test too 😅 )

Quetzacoalt91: When using trans() with a domain, both are valid.  :)

src/content/1.7/modules/creation/module-translation/new-system.md
```diff
@@ -140,11 +140,11 @@ Since the module is called MyModule, the translation domain should be `Modules.M
 ```php
 // file: controllers/front/something.php
 
-class MyModuleSomethingFrontController extends ModuleFrontController
+class MySomethingModuleFrontController extends ModuleFrontController
```
eternoendless: This is wrong. If the module is called `mymodule` then the controller `Something` must be called like this:

```suggestion
class MymoduleSomethingModuleFrontController extends ModuleFrontController
```

> the class name follows this format: `<ModuleName><FileName>ModuleFrontController`.

[source](https://devdocs.prestashop.com/1.7/modules/concepts/controllers/front-controllers/)

src/content/1.7/modules/creation/module-translation/new-system.md
matks: @Quetzacoalt91 If I try to use `$this->module->trans()` in a module front controller with PS 1.7.7, I get the following error:
Capture d’écran 2019-12-30 à 12 06 12

src/content/1.7/modules/creation/module-translation/new-system.md
matks: Indeed, bad copy-paste

matks:
Thank you for review

PR #432, opened 2019-12-31
Create dedicated page for webservice access

### LINKED ISSUE #431: How to: Create Webservice credentials programmatically
Linked to #342.

Contributors have asked several times how to create PS webservice credential via their module, for instance at installation.

As far as I know there is some code to write to make it possible, this should be described in the doc.

Quetzacoalt91:
Fixes #431 

This PR moves the creation of Webservice accounts in a dedicated page of the doc, then adds content to do it with PHP code.


matks:
Thank you @Quetzacoalt91

PR #435, opened 2020-01-06
Update alert block for composer dependencies

Quetzacoalt91:


src/content/1.7/modules/concepts/composer.md
```diff
@@ -111,7 +111,8 @@ correctly you **must not** override its dependencies. Which is why you need to a
 
 **Don't forget to publish your vendor folder**
 
-When using composer the autoload file generated by composer is required for your module to work correctly. So before
-creating your module archive and releasing it don't forget to run `composer dump-autoload -o` (or `composer install --no-dev` if
-you have included production dependencies) so that they will be included in your module.
+Composer is responsible in downloading your dependencies and generating the autoloader from the `composer.json`
+& `composer.lock` in your source code. These two artifacts are needed by your module to work properly, so before
+releasing it don't forget to run `composer dump-autoload -o --no-dev` to include your **production dependencies**
```
PierreRambaud:
```suggestion
releasing it don't forget to run `composer dump-autoload -o --no-dev` to only include your **production dependencies**
```

Quetzacoalt91: There is something weird about this `only`. I wrote it at the beginning but decided to remove it, it's like we add the prod dependencies but we forget all the rest

src/content/1.7/modules/concepts/composer.md
```diff
@@ -111,7 +111,8 @@ correctly you **must not** override its dependencies. Which is why you need to a
 
 **Don't forget to publish your vendor folder**
 
-When using composer the autoload file generated by composer is required for your module to work correctly. So before
-creating your module archive and releasing it don't forget to run `composer dump-autoload -o` (or `composer install --no-dev` if
-you have included production dependencies) so that they will be included in your module.
+Composer is responsible in downloading your dependencies and generating the autoloader from the `composer.json`
```
eternoendless:
```suggestion
Composer is responsible for downloading your dependencies and generating the autoloader from the `composer.json`
```

src/content/1.7/modules/concepts/composer.md
```diff
@@ -111,7 +111,8 @@ correctly you **must not** override its dependencies. Which is why you need to a
...
+& `composer.lock` in your source code. These two artifacts are needed by your module to work properly, so before
+releasing it don't forget to run `composer dump-autoload -o --no-dev` to include your **production dependencies**
+in the archive.
```
eternoendless: I think this notice is confusing.

Here's my suggestion:

```
{{% notice warning %}}
**Don't forget to include your vendor folder in your release package**
Composer is responsible for downloading your dependencies and creating the autoloader file that your module will need to work properly. Therefore, remember to run `composer dump-autoload -o --no-dev` before you create your package, and make sure you include the `vendor` directory in it.
{{% /notice %}}

{{% notice warning %}}
**Don't include development dependencies in your release package**
Development-only libraries like PHPUnit can pose a security threat in production environments. Make sure that these libraries are imported as "dev dependencies" (`require-dev`) so that they aren't included when you create your release package. If in doubt, **double-check that they aren't  in the `vendor` directory before creating your release package**.
{{% /notice %}}
```

src/content/1.7/modules/concepts/composer.md
PierreRambaud:
```suggestion
releasing it don't forget to run `composer dump-autoload -o --no-dev` to install your dependencies for a **production** environment
```
Maybe better? wdyt?

src/content/1.7/modules/concepts/composer.md
Quetzacoalt91: > double-check that they aren't in the `vendor` directory before creating your release package

Should be `before submitting` or `when creating` don't you think?

Quetzacoalt91:
Doc updated with your recommandations

eternoendless:
Thank you @Quetzacoalt91

PR #442, opened 2020-01-14
Fix typo in tabs.md

### LINKED ISSUE #437: Method name issue
There is an issue in https://github.com/PrestaShop/docs/blob/master/src/content/1.7/modules/concepts/controllers/admin-controllers/tabs.md

We say we can use `Link::getAdminkLink('MyModuleDemoController')'` to retrieve the link from the controller. 
`getAdminkLink` should be replaced by `getAdminLink`

Quetzacoalt91:
Fixes #437


PierreRambaud:
Thanks @Quetzacoalt91

PR #453, opened 2020-01-28
Update package.json for security reason

PierreRambaud:
Webpack must be update from version 2 to 4
extract-text-webpack-plugin is replaced by mini-css-extract-plugin
css-loader from 0.28 to 3.4

Tests:
Run the documentation and test that everything are still working fine.
Change a css line and make sure it's properly executed when running in dev / prod mode.


eternoendless:
You don't need to commit the `main.js` file, it's a side effect of using webpack to compile scss

PierreRambaud:
@eternoendless Done, I also add them into the `.gitignore` file

eternoendless:
Thank you @PierreRambaud

PR #454, opened 2020-02-11
Fix formatting .`tpl` -> `.tpl`

DrevanTonder:
Saw that formatting issue while I was reading the docs. I'm quite sure .`tpl` should be `.tpl`


matks:
Nice catch @DrevanTonder thanks !

PR #461, opened 2020-02-18
Add a paragraph about compatible licenses

ttoine:
next step should be to add a paragraph about technical requirements

### DIFF
--- a/src/content/1.7/contribute/contribution-guidelines/_index.md
+++ b/src/content/1.7/contribute/contribution-guidelines/_index.md
@@ -164,6 +164,33 @@ Remember, this is only needed for the Pull Request form, not for your commit mes
 
 [how-to-compile-assets]: {{% ref "1.7/development/compile-assets.md" %}}
 
+
+## Dependencies
+
+Adding third party software in the core or in a module might sometimes be faster and easier than to develop it from scratch and then to maintain it. Composer and NPM are used to manage such dependencies in the PrestaShop open source project.
+
+It is possible to add new dependencies, after a careful selection and study. A few rules must be followed.
+
+### Licenses
+
+The first step is to check the **legal compliance** of a third party software. The PrestaShop open source project being distributed under the OSL license, dependencies must be compatible with it. 
+
+Below is a list of the known compatible licenses that can accepted:
+
+- MIT
+- ISC
+- BSD
+- AFL
+- EUPL
+- Apache 
+
+Then, a last check is done to be sure that there is no additional clause that would introduce a restriction of use.
+
+If the legal compliance is confirmed, it will be possible to go the next step: the technical review.
+
+
 ## Read more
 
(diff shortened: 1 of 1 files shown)


Quetzacoalt91:
Hi @ttoine,

There is a list of accepted licenses in https://devdocs.prestashop.com/1.7/modules/sell/techvalidation-checklist/#license-is-compatible. To avoid duplicated / obsolete content, what should be updated on the other page?

ttoine:
Good point, thanks.

You can add Apache to the list on this page, and remove OSL from the list on the "module page". Currently, modules must be in AFL, not in OSL. And while this is possible to add AFL code in OSL, the opposite is not really possible.
Also note that this list is important for native modules, but also for third party modules made available on Addons

I will add CC licenses for the artwork stuff in the "contributing guidelines" page, this is missing.

ttoine:
I have added a few other licenses, regarding code and artwork.
@Quetzacoalt91 I think that might be interesting.

@PierreRambaud what should be removed ? the legal compliance of the core, and of a module is not exactly the same (at least, OSL code is not really possible in modules)

src/content/1.7/contribute/contribution-guidelines/_index.md
```diff
@@ -173,18 +173,31 @@ It is possible to add new dependencies, after a careful selection and study. A f
...
+Additionally, artwork (e.g: icons, pictures, fonts, but not only) is usually distributed with specific licenses. Below is a list of known compatible licenses that can be accepted for artwork:
+
+- CC-0
+- CC-by-sa
+- CC-by
+- Art Libre
+- Artistic
+- Unlicense
+
+In general, public domain is not acceptable, as it doesn't exist in all juridictions or countries.
```
eternoendless: Are you certain about this? We are currently using these flags for example, which don't specify one of the aforementioned licenses: http://www.famfamfam.com/lab/icons/flags/

ttoine: certain about what ? artistic / unlicense? or public domain? or both?

ttoine: I checked famfamfam icons pack. There is a readme that give more information: they "are free for any use (attribution appreciated but not required)"

It would be better if they could be in CC-0, as this license includes a workaround for countries where public domain does not exist.

ttoine: Additional information: just keep in mind that open source licenses are used to give right to the users and protect them from when they use the software (use, study, modify, redistribute)

If there is nothing explained in a package regarding the protection of users and what they can do and not do, then usually, both FSF and OSI recommendation is to avoid it.

PierreRambaud:
Thanks @ttoine

ttoine:
yw :-)

Issue #464, opened 2020-02-20
FAQ on shipping informations


toutantic:
Question in need for an answer that should go here:
https://devdocs.prestashop.com/1.7/faq/shipping/

## Shipping information
Where and how should I read and write the following informations to be sure that my module is compatible with every shipping modules:
- Tracking number
- Carrier name
- Estimated delivery date
- Real delivery date

PR #467, opened 2020-02-20
Add answer for first shipping question in FAQ

Quetzacoalt91:
Fixes https://github.com/PrestaShop/docs/issues/464

### DIFF
--- a/src/content/1.7/faq/shipping.md
+++ b/src/content/1.7/faq/shipping.md
@@ -10,14 +10,48 @@ title: Shipping FAQ
 
 ## Shipping information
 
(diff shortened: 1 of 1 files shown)


toutantic:
Thank you @Quetzacoalt91

toutantic:
@PrestaShop/prestashop-product-team We have a use case for a module that needs the real delivery date of an order (insurrance on delivery date...). This field does not exist in the core,should we add it ?

colinegin:
Isn't it the date where the order has had the status "Delivered" ?
Do you have more details about this use case so that I understand it clearly ?

PR #474, opened 2020-02-28
Update theme translation documentation

eternoendless:
Update guide on how to translate a theme, using internal documentation (in French) as basis.


matks:
Thank you @eternoendless

PR #475, opened 2020-02-29
Remove unnecessary argument

Arman-Hosseini:
The default value of $active parameter is true


matks:
> @matks Please check out this.

Thank you for the nice catch !

matks:
Thank you @Arman-Hosseini

PR #481, opened 2020-03-06
Demo View Order hooks module - one more bug found and some docs added

tdavidsonas88:


matks:
Thanks @tdavidsonas88

PR #485, opened 2020-03-15
Update license header year in example files

Arman-Hosseini:


matks:
Thanks @Arman-Hosseini !

PR #486, opened 2020-03-16
Replace http links by https

matks:


matks:
Merging without review because https > http

PR #491, opened 2020-03-20
Improve notice tip to cloning repository

presteus:
Improve notice tip to cloning repository with git configuration for case-sensitive systems.

### DIFF
--- a/src/content/1.7/basics/installation/_index.md
+++ b/src/content/1.7/basics/installation/_index.md
@@ -105,6 +105,11 @@ PrestaShop comes in two "flavors":
 **Prefer cloning the repository using git for the development version.**
 
 If you intend to work on PrestaShop itself, we suggest using Git to clone the source code of PrestaShop from the GitHub repository.
+
+**For case-sensitive systems, you need to configure git to ignore case:**
+```shell
+git config --global core.ignorecase true
+```
 {{% /notice %}}
 
 {{% callout %}}


PierreRambaud:
I have a case sensitive system and never had to use this option, I'm interesting about when it's needed, could you give me an example?

matks:
@PierreRambaud see https://github.com/PrestaShop/PrestaShop/issues/18247 😉

PierreRambaud:
> @PierreRambaud see [PrestaShop/PrestaShop#18247](https://github.com/PrestaShop/PrestaShop/issues/18247)

Weird, I'm on the same system and never had this kind of errors :sweat_smile: 
I found the real reason, it's not because of case-sensitive system, it's because of filesystem you're using like APFS, HFS+, FAT, NTFS.

Could you please change the wording to talk about filesystem instead of system :)

presteus:
> 
> 
> > @PierreRambaud see [PrestaShop/PrestaShop#18247](https://github.com/PrestaShop/PrestaShop/issues/18247)
> 
> Weird, I'm on the same system and never had this kind of errors 😅
> I found the real reason, it's not because of case-sensitive system, it's because of filesystem you're using like APFS, HFS+, FAT, NTFS.
> 
> Could you please change the wording to talk about filesystem instead of system :)

You are right, i changed **system** word to **filesystem.
FYI: I use ext4 filesystem.

PierreRambaud:
> > > @PierreRambaud see [PrestaShop/PrestaShop#18247](https://github.com/PrestaShop/PrestaShop/issues/18247)
> > 
> > 
> > Weird, I'm on the same system and never had this kind of errors sweat_smile
> > I found the real reason, it's not because of case-sensitive system, it's because of filesystem you're using like APFS, HFS+, FAT, NTFS.
> > Could you please change the wording to talk about filesystem instead of system :)
> 
> You are right, i changed **system** word to **filesystem.
> FYI: I use ext4 filesystem.

I'm also on ext4 :scream:

Progi1984:
Ping @presteus

matks:
Hi @presteus, as we are not 100% sure of this one I prefer to close the PR. We dont want to write something that might be false for some people ; there is nothing worse than a wrong documentation 😅 

Thank you for your contribution

PR #496, opened 2020-03-29
Minor typo on Embedding a template in the theme section

dvdwalker:
On the example "public function hookDisplayLeftColumn($params)" there was a square bracket closing the smarty-assing instruction and should be a parenthesis


PierreRambaud:
Thanks @dvdwalker

PR #502, opened 2020-04-08
How to work with grid search form

### LINKED ISSUE #497: Write documentation about filtering grids
https://devdocs.prestashop.com/1.7/development/components/grid/tutorials/work-with-search-form/

This page was left blank and should be documented. It should be added in https://devdocs.prestashop.com/1.7/development/components/grid/#search-criteria Which talks about Filters component but no real link associated

This feature has evolved along the versions so it's important to detail the solutions for the different versions

Content to add in the doc:
- Definition::getFilters (which is the core o the feature, including for build the form in the grid)
- Filters class (since 1.7.5)
- SearchParametersResolver (automatic resolving in 175, explain limitations - double grid)
- FiltersBuilderResolver (automatic resolving since 176)
- Different filtering possibilities
  - controller/action for version previous to 176
  - filterId / GRID_ID + ResponseBuilder (since 1.7.6)
  - interface IdentifiableGridDefinitionFactoryInterface and use of CommonController (since 1.7.7)
- Thorough explanation of `SearchAndResetType` along with the different clean methods/routes

jolelievre:
- Add page to learn how to work with grid search form
  - Special page for 1.7.5
- Details on filter form types
- Create your custom form factory


eternoendless:
Hi @jolelievre 

Thank you for this initiative. However, I see some problems with this kind of feature:

1. **It's not search friendly.**
	Search engines will see the whole page and won't understand that different sections belong to different versions.

2. **The selector will cause problems when we'll release a new major version.** 	
	You may have noticed that all paths start with `/1.7/`. Same as in the symfony documentation, my idea was that when we'll release a new major version, we can create a new path like `/8.0/`, copy everything inside, and work from there, then switch between versions with a selector like the one you did. (Note: [Symfony docs actually has one branch per version](https://github.com/symfony/symfony-docs), I'm not sure how they get to integrate them all in a single site afterwards.)

For now, I think it would be simpler to write different sub articles for each version if they diverge, like we did here: https://devdocs.prestashop.com/1.7/modules/creation/module-translation/

jolelievre:
Hi @eternoendless 

Indeed I didn't have notice but in Symfony doc each version has its dedicated page:
- https://symfony.com/doc/3.4/forms.html
- https://symfony.com/doc/3.2/forms.html

I'd like to keep the page readable, because each version has a lot of differences and having all the versions described in the same page will make it really "heavy".
Maybe we could split it into three different pages?

- src/content/1.7/development/components/grid/tutorials/1.7.5/work-with-search-form.md
- src/content/1.7/development/components/grid/tutorials/1.7.6/work-with-search-form.md
- src/content/1.7/development/components/grid/tutorials/work-with-search-form.md (which would be for 1.7.7)

And I could modify the selector so that it's actually some links to the other versions? (seems like it's the case in Symfony doc as well). This is also SEO friendly.

This would solve problem 1. And for 2. when we will have a new major 8.0 we don't need to keep the previous versions, only the latest one (whether we deprecate the other features or not).

eternoendless:
We agree, only with different words ;)

jolelievre:
@eternoendless As we discussed I separated into two pages:
- the generic one for 1.7.6 + 1.7.7
- an additional one for 1.7.5

I also added a page to create a custom form grid factory

@mickaelandrieu I would be curious to have you feedback on this doc

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
```diff
@@ -0,0 +1,146 @@
...
+            - '@prestashop.core.grid.definition.factory.category'
+            - '@prestashop.core.grid.data.factory.category_decorator'
+            - '@prestashop.core.grid.filter.category_form_factory'
+            - '@prestashop.core.hook.dispatcher'
+```
+
+## Simply change the action
+{{< minver v="1.7.7" title="true" >}}
+
+If you just need to change the form action PrestaShop provides a `FilterFormFactoryFormActionDecorator` decorator that simply sets the action you need to use. All you need to do is define the services properly:
```
mickaelandrieu, line 127:
```diff
- simply
+ 
```

Always try to avoid "simply", "easily", "obviously"... I used to do the same mistake, and this make the reader feel (s)he's bad/stupid for free. Let's assume all topics are difficult for people who don't master it, yet :)

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
```diff
@@ -0,0 +1,146 @@
...
+
+```yaml
```
mickaelandrieu, line 129:
```diff
+ # your-module/config/services.yml
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/in-1-7-5.md
```diff
@@ -0,0 +1,237 @@
...
+- `Filters` class
+- Controllers:
+  - Search action
+  - List action
+  - Reset filter action
+
+## Add Filters into GridDefinition
+
+```php
+final class EmailLogsDefinitionFactory extends AbstractGridDefinitionFactory {
```
mickaelandrieu, line 21: This doesnt bother me that much, but your bracket should be positionned in the next line :)

mickaelandrieu (changes requested):
Also, why not talking about developer tools module ? It provides a lot of useful information about [grids](https://github.com/friends-of-presta/developer_tools)

src/content/1.7/development/components/grid/filter-types-reference/_index.md
```diff
@@ -0,0 +1,158 @@
...
+
+### CountryChoiceType
+
+Builds a country selector, it contains various options but they are automatically filled with the list of country.
+
+| Properties         | Expected value(s)             |
+|--------------------| ------------------------------|
+| **Options**        | `choices`                     |
+|                    | `choice_attr`                 |
+|                    | `withDniAttr`                 |
```
jolelievre, line 38: This shouldn't use camel case notation The PR to fix this in the core is here https://github.com/PrestaShop/PrestaShop/pull/18686
And the doc must be updated accordingly
Thanks @PierreRambaud for spotting this issue

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,290 @@
...
+  defaults:
+    _controller: 'PrestaShopBundle:Admin/Sell/Catalog/Manufacturer:search'
+    _legacy_controller: AdminManufacturers
+    _legacy_link: AdminManufacturers:submitFiltermanufacturer
+```
+
+#### Common search controller
+{{< minver v="1.7.7" title="true" >}}
+
+As this controller is almost always the same, starting 1.7.7, we introduced a common controller. So all you need to do is define the routing:
```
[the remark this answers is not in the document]
jolelievre: It's true that the tag is present, but I want to emphasize that this is only possible after 1.7.7
You think it's too much?

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,290 @@
...
+                    'redirect_route' => 'admin_manufacturers_index',
+                ])
+                ->setAssociatedColumn('actions')
+            )
+    ...
+```
+
+## Search form in 1.7.5
+
+In the 1.7.5 version, first one that integrated the grid component, the system was a bit different and used controller/action to identify the filters. It had many drawbacks and was abandoned in favor of a unique id but in case you need it here is [how to work with search form in 1.7.5]({{< ref "/1.7/development/components/grid/tutorials/work-with-search-form/in-1-7-5.md" >}}).
```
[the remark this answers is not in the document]
jolelievre: You're right ^^ I'll replace with *a few drawbacks*

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
```diff
@@ -0,0 +1,146 @@
...
+    ) {
+        $this->formFactory = $formFactory;
+        $this->urlGenerator = $urlGenerator;
+        $this->requestStack = $requestStack;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create(GridDefinitionInterface $definition)
```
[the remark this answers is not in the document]
jolelievre: I'll remove the form changes and only leave the url changes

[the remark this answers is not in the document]
jolelievre: Actually not that simple the form builder is recreated based on the generated form, because the setAction can't be called on the form itself..

src/content/1.7/development/components/grid/tutorials/work-with-search-form/in-1-7-5.md
jolelievre: Yep nice catch

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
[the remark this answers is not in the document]
jolelievre: Looking at this in more details, there's actually few simple things you can do by decoration this form factory because you only have access to the GridDefinition, and the generated form:
- modifying the definition could be done here as a simple example, but it's not the correct way to do it since we already have the `*GridDefinitionModifier` that is made for that
- modifying the generated form is not possible, the `FormInterface` nearly only has getters, we could add some extra form fields, but here again it should be handled via modifying the definition filters

What would have been nice is to be able to get the form builder from the core factory, but the only function in `GridFilterFormFactoryInterface` is `create` and we can't modify it to add a `getFormBuilder` function (this would have made the action change much easier).

So in any was decorating this service will always be used for advanced modification, so having an advanced example makes sense.

Unless you have an idea of a simple use case, in which case we will replace this one, I'd say we keep this one for now.

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+  - Search action
+  - List action
+  - Reset filter action
+
+## Add Filters into GridDefinition
+
+You should use `AbstractGridDefinitionFactory` as a parent class, and define a const `GRID_ID` which will be used as a key to persist the `Filters`.
+
+```php
+final class ManufacturerGridDefinitionFactory extends AbstractGridDefinitionFactory {
```
Progi1984, line 24:
```suggestion
final class ManufacturerGridDefinitionFactory extends AbstractGridDefinitionFactory
{
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+            )
+        ;
+    }
+    ...
+}
+```
+
+### The filters types
+
+In the filters collection you define all the available filters (which will match your grid columns), you can define a specific type depending on the column.
```
Progi1984, line 82:
```suggestion
The filters collection permits to define all the available filters (which will match your grid columns). You can define a specific type depending on the column.
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+You can basically use any Symfony form type (including your custom ones) and PrestaShop provides a few [filter types]({{< ref "/1.7/development/components/grid/filter-types-reference/_index.md" >}}) that might be useful to you.
+
+### Filterable grid definition
+{{< minver v="1.7.7" title="true" >}}
+
+Starting the 1.7.7 we made it even easier, you can use the `AbstractFilterableGridDefinitionFactory`. This will allow you use the [common search controller](#common-search-controller).
+
+```php
+final class AddressGridDefinitionFactory extends AbstractFilterableGridDefinitionFactory {
```
Progi1984, line 91:
```suggestion
final class ManufacturerGridDefinitionFactory extends AbstractFilterableGridDefinitionFactory {
```

That permits to have a sample who follow the page.

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+    const GRID_ID = 'address';
```
Progi1984, line 92:
```suggestion
    const GRID_ID = 'manufacturer';
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+}
+```
+
+## Controller actions
+
+The Grid filtering workflow is divided into three actions:
+
+- search action: it parses the filters from the POST request, then redirects to the list action
+- list action: it parses the filters from GET request, persists them into database and finally renders the grid
+- reset action: to clean the persisted filters and reset to the default ones
```
Progi1984, line 144:
```suggestion
- reset action: it cleans the persisted filters and reset to the default ones
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+
+{{% notice info %}}
+In this tutorial we assume the search and list actions have the same url, thus we don't need to specify the search route in the form action. Search manages the **POST** request and list the **GET** request.
+
```
Progi1984, line 148:
```suggestion
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+            'manufacturerGrid' => $this->presentGrid($manufacturerGrid),
+        ]);
+    }
+    ...
+}
+```
+
+### Reset action
+
+Last action is used to reset the persisted filters and your grid filtering/sorting. This action is the same for nearly all grids so PrestaShop provides a common controller to manage it, and you actually already set it via the grid definition.
```
Progi1984, line 271:
```suggestion
This action is used to reset the persisted filters and your grid filtering/sorting. This action is the same for nearly all grids so PrestaShop provides a common controller to manage it, and you actually already set it via the grid definition.
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
```diff
@@ -0,0 +1,293 @@
...
+
+It is defined in the `SearchAndResetType` options, it uses `admin_common_reset_search_by_filter_id` and only needs the filter id to identify the filters to clear, and a redirection route.
```
Progi1984, line 273: For `SearchAndResetType`, add a link to the first file. Improve the internal linking.

src/content/1.7/development/components/grid/tutorials/work-with-search-form/_index.md
jolelievre: This is on purpose, because we don't explain the same way: `AbstractFilterableGridDefinitionFactory` vs `AbstractGridDefinitionFactory` which is only possible in 177

And it matches the current grid definitions in the core which are also different

Progi1984: Ok. Misread

Progi1984 (approved):

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
```diff
@@ -0,0 +1,148 @@
+---
+title: Create your custom form factory
+menuTitle: Custom form factory
+weight: 1
+---
+
+# Create your custom form factory
+
+In some cases the generic form factory is not enough for your need. In PrestaShop the category grid has two specific features:
```
atomiix, line 9:
```suggestion
In some cases the generic form factory is not enough for your need. In PrestaShop the category grid has three specific features:
```

src/content/1.7/development/components/grid/tutorials/work-with-search-form/custom-form-factory.md
```diff
@@ -0,0 +1,148 @@
...
+- it can manage a parent category which changes the grid rendering (only displays children) so it needs to dynamically add a parameter
+- its search url doesn't match the list so it needs to specify the form action
+- the redirection url is dynamic so that it takes the parent category into account 
+
+To manage this additional behaviour we create a decorator that uses the default form factory but adds specific behaviour.
+
+## Decorate the default factory
+
+{{% notice note %}}
+In those example we use decorator that are based on the default grid factory. But you could create a fully autonomous class as long as it implements the `GridFilterFormFactoryInterface`.
```
atomiix, line 20:
```suggestion
In those examples we use decorator that are based on the default grid factory. But you could create a fully autonomous class as long as it implements the `GridFilterFormFactoryInterface`.
```

atomiix (dismissed):

jolelievre:
@atomiix All done ^^

PierreRambaud (approved):

Progi1984 (approved):

atomiix (approved):

PR #504, opened 2020-04-16
Add ajax parameter for ajax call

luigimassa:
I added a small specification when the call of the controller is done via ajax


src/content/1.7/modules/concepts/controllers/front-controllers.md
```diff
@@ -166,6 +166,17 @@ secured by redirecting it to HTTPS.
...
+
+When you call controller via ajax, you nedd to add ajax parameter in teh url.
+
+```php
+Context::getContext()->link->getModuleLink('cheque', 'validation', array('idPayment' => 1337, 'ajax'=>true);
+```
+
+* Without URL rewriting: `http://<shop_domain>/index.php?idPayment=1337&fc=module&module=cheque&controller=validation&id_lang=1&ajax=true`
+* With URL rewriting: `http://<shop_domain>/en/module/cheque/validation?idPayment=1337&ajax=true`
+
```
Quetzacoalt91, line 179: What do you think about moving this part between `### Example of method calls` and `## Restricting access`? We could bring these examples together.

luigimassa: I think it's a good idea. In ### Example of method calls is best.

Progi1984:
Ping @luigimassa

src/content/1.7/modules/concepts/controllers/front-controllers.md
```diff
@@ -137,6 +137,17 @@ Context::getContext()->link->getModuleLink('cheque', 'validation', array('idPaym
 * Without URL rewriting: `http://<shop_domain>/index.php?idPayment=1337&fc=module&module=cheque&controller=validation&id_lang=1`
 * With URL rewriting: `http://<shop_domain>/en/module/cheque/validation?idPayment=1337`
 
+### Ajax request
+
+When you call controller via ajax, you nedd to add ajax parameter in teh url.
```
Progi1984, line 142:
```suggestion
When you call controller via AJAX, you nedd to add `ajax` parameter in the url.
```

matks:
Thank you @luigimassa

PR #515, opened 2020-04-27
Webservers

djodjo3:
Starting to discuss about webservers in the devdocs.
Won't be able to add everything at once, but that's a start.
Let me know what you think is missing, should be reworded, etc..


src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+- 1.7/optimizations/
+---
+
+# Optimize your webserver
+
+We often talk about optimizing the application side but less about the webserver.
+
+Which is a shame as there is a lot to gain from some simple configurations parameters. And you don't need to have hundred of visitors for them to be efficient, either to reduce your server's load or speed up your visitor's browsing.
+
+As we'll see, the configuration options we'll tweak are quite easy and should allow some nice improvements.
```
matks, line 13: On devdocs we try to have a neutral and "official" tone so we avoid shortcuts 😉
```suggestion
As we will see, the configuration options we'll tweak are quite easy and should allow some nice improvements.
```

Same for "don't need" => "do not need", and this kind of things 😄 we try to look professionnal

src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+
+# Webserver's function
+
+You may already be aware of, a webserver is a piece of software dedicated to satisfy world wide web requests.
```
matks, line 17: Feel free to link to external resources for people who would like to discover this topic or learn more.
For example here you can link to https://en.wikipedia.org/wiki/Web_server

src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+
+It usually works through the http (and https) protocol and delivers web pages to the clients (your web browser). 
+
+{{< figure src="../img/architecture.png" title="webserver's function" >}}
+
+An interesting point to mention is that, php being compiled server side, it often is the webserver's job to manage this task before sending html to the web browser.
+
+## Testing your optimizations
+
+There are plenty of ways to validate the gains of those improvements but a very simple one is to go check any page speed website.
```
matks, line 27: When I see this 😄I think "great ! let's try !" and I look for ... links to tools or services to do them. Could you add one or two speed check ?

src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+As usual, the surest way to tune your shop is:
+
+- benchmark
+- make a change
+- repeat
+{{% /notice %}}
+
+### Webservers
+
+We'll show you the different ways to implement those options for both nginx and httpd, the two most popular and widespread webservers.
```
matks, line 41:
```suggestion
We'll show you the different ways to implement those options for both [nginx](https://www.nginx.com/) and [httpd](https://httpd.apache.org/docs/2.4/programs/httpd.html), the two most popular and widespread webservers.
```

Linking example 😉

src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+
+Apache's httpd is the usual suspect and the most popular webserver of alls.
+
+Contrary to nginx, it uses threads that directly serves the contents, being locked during the process.
+
+Which means that if the server receives more calls than available threads it won't be able to answer - hence the reputation of being able to handle less traffic than nginx.
+
+### Optimizations
+
+And now to the crunchy part !
```
matks, line 60: Maybe something more neutral ? 😄 

```suggestion
Now that you know better httpd and nginx, we can see how to improve their performance.
```

src/content/1.7/scale/webserver.md
```diff
@@ -0,0 +1,169 @@
...
+
+And here is the way to enable gzip compression to your httpd configuration:
+
+First, make sure the deflate module is activated in your httpd configuration file:
+
+```
+LoadModule deflate_module modules/mod_deflate.so
+```
+
+Then, you can add the following lines to your VirtualHost:
```
matks, line 95:
```suggestion
Then, you can add the following lines to your VirtualHost:
```
I believe this can be done in other configuration files, right 🤔 

Remember : when you write doc, some people (junior developers or junior ops) believe What You Say Is the Truth. So if you say "put it in this file", they will think it cannot go somewhere else. You might want to say you suggest putting it in this file as this is the classical way to do it, but mention other possible areas.

toutantic:
nginx and apache recommandations should be on separate pages.
People reading this will look for specific instruction about the webserver they are working with and won't be interested in the other.

toutantic:
I believe this content could be greatly improved by following those guildelines:
https://documentation.divio.com/introduction/

The current document is a mix of explanation, how-to and reference. You should choose one type of documentation.
I believe this one should be reference or how-to but not explanations.

src/content/1.7/scale/webserver.md
djodjo3: I did think about it but I did not want to influence anyone using any specific tool.

src/content/1.7/scale/webserver.md
djodjo3: Tried to rework that in new version, let me know what you think ;)

src/content/1.7/scale/webserver.md
djodjo3: Added link.

src/content/1.7/scale/webserver.md
djodjo3: Still, added a google search link, let me know what you think of it :p

src/content/1.7/scale/webserver.md
djodjo3: Added links 👍

src/content/1.7/scale/webserver.md
djodjo3: Removed this part in new version, but crunchy ain't neutral ? 😇

src/content/1.7/scale/webserver.md
djodjo3: To be clear, VirtualHost is not really a file but a configuration section (in a file, obviously ^^).
I tried to be more explicit in the new version, but paths and filenames may vary depending on versions and distribution and OS.
So, it's a bit hard to say, for sure: "add this in this file, at this path".
Maybe "supporting" some OS and version would be a solution.

djodjo3:
So, I've reworked this PR as follow:
* Moved the more reference/how-to parts to dedicated sections for both nginx *and* httpd
* Moved the webservers explanation part as an introduction section, so it *should* not clash between the functions

I do hope it works, but let me know what you think.

Once we have the correct structure, we should be able to add up more materials more easily.

toutantic:
You got it, this separation make this documentation much easier to read.
People who already know the concepts can skip directly to the how-to part, people who don't know still have the explanations !

toutantic:
May be replace httpd with httpd/apache2
I believe people will search more for apache than httpd ?

matks:
> May be replace httpd with httpd/apache2
> I believe people will search more for apache than httpd ?

Very SEO oriented 😄 

I will fix it

matks:
See https://github.com/PrestaShop/docs/pull/521 I did not replace all mentions but I inserted it in main titles for SEO to be optimized

PR #522, opened 2020-05-01
Missing use of FolderThemeScanner

Leenwt:
Added the use case for FolderThemeScanner which threw an error for me


PierreRambaud:
Thanks @Leenwt

PR #532, opened 2020-05-06
Provide an example for cron tasks in modules

Quetzacoalt91:


src/content/1.7/modules/concepts/controllers/front-controllers.md
```diff
@@ -186,3 +186,50 @@ public $ssl = true;
   * **initContent()**: Initializes the content.
   * **initFooter()**: Called after initContent().
   * **display()** or **displayAjax()**: Displays the content.
+
+## Using a front controller as a cron task
```
PierreRambaud, line 190: Not sure it's a good idea to use front controllers.
If someone has the link, he can spam it :thinking:

Quetzacoalt91: What would be your solution?

AFAIK this is the only way we know to run code in a PrestaShop context, with all the core classes instancied. I can add a check to we exit if PHP is not run via CLI.

PierreRambaud: Create script in admin directory with custom token as protection:
https://github.com/PrestaShop/ps_facetedsearch/blob/master/ps_facetedsearch-attribute-indexer.php
It's not the cleaner way but cron scripts must be protected to prevent spamming :/

Quetzacoalt91: That's actually what I'm trying to avoid : Having this kind of PHP you provide file won't load totally the core of PrestaShop. If you remember, from PS 1.7.6 modules had to replace these files in favor of Controllers to make sure all the Context was initiated (otherwise displaying a price was failing by triggering PHP warnings).

The protection you're talking about is missing from these examples (no call to `php_sapi_name()`, no token check...), I agree with you.

PierreRambaud: Add these checks and everything will be better. But in the future we should find a way to have unauthenticated admin modules controllers, wdyt?

Quetzacoalt91: I can't agree more. Commands are already a solution for this need, but we can't fully rely on them when a module is compatible with all PS 1.7 versions.

Quetzacoalt91: PR has been updated accordingly

Progi1984:
Thanks @Quetzacoalt91

PR #542, opened 2020-05-25
Add notice warning for webservice usage

matks:
Struggled a bit this morning as my httpd was ignoring the `.htaccess` so I decided to improve the documentation


matks:
Thanks for review !

PR #544, opened 2020-05-26
Product form documentation

jolelievre:
Document all the product form details, for old version 

This is a WIP but it could be merged as is and udpated during development

- needs minver details for required fields in old form
- new version (split work in progress)
  - define all form types
  - define associated commands
  - add some screenshots of the page with borders to show where each form type is presented


src/content/1.7/development/page-reference/back-office/product/form/old-product-form.md
```diff
@@ -0,0 +1,198 @@
+---
+title: The old Product Form
+menuTitle: Old Form
+weight: 1
+---
+
+# The Old Product Form
+{{< minver v="1.7.0" title="true" >}}
```
eternoendless, line 8: This is only needed when something is introduced in a minor (or patch) version. The old form was introduced in 1.7.0 so not needed here. 

You should clarify at the top of the other page that the form is being remade for 1.7.8 to be coherent with the rest of the migration and that for versions 1.7.0~1.7.7 people should read this article.

That said, maybe this page should be called something like "Product form before 1.7.8".

jolelievre: Good idea, done!

src/content/1.7/development/page-reference/back-office/product/form/_index.md
```diff
@@ -0,0 +1,254 @@
+---
+title: The Product Form
+menuTitle: Product Form
+weight: 42
```
eternoendless, line 4: Can be removed
```suggestion
```

src/content/1.7/development/page-reference/back-office/product/form/_index.md
```diff
@@ -0,0 +1,254 @@
...
+---
+
+# The New Product Form
+{{< minver v="1.7.8" title="true" >}}
+
+In most recent Product page, which was remade in 1.7.8, the form is divided into multiple sub forms. Each form has dedicated CQRS commands to manage its content.
+For previous versions you can read this documentation about [product form before 1.7.8]({{< ref "1.7/development/page-reference/back-office/product/form/product-form-before-1-7-8" >}}).
```
eternoendless, line 11: Consider adding this in a notice tip like here: https://devdocs.prestashop.com/1.7/modules/creation/module-translation/new-system/

src/content/1.7/development/page-reference/back-office/product/form/_index.md
```diff
@@ -0,0 +1,254 @@
...
+| **ProductCategoriesType** | `product_categories` | Manage product categories | N/A |
+
+### BasicInformationType
+
+| Fields | Field type                       | Description            |
+|:-------|:---------------------------------|:-----------------------|
+| `product_id` | `HiddenType` | The *Product* ID |
+| `type_product` | `ChoiceType` | Type of product: Standard, virtual or Pack of Product |
+| `name` | `TranslateType` | Product name (localized) |
+| `active` | `CheckboxType` | Boolean to indicate if Product is `active` |
```
zuk3975, line 44: as we discussed this will not have active field, as it will be disabled by default after creating a ghost product.

src/content/1.7/development/page-reference/back-office/product/form/_index.md
```diff
@@ -0,0 +1,254 @@
...
+| **RelationshipsType**    | `relationships` | Contains relationships of the product | N/A |
+| **ShortcutType**         | `shortcuts`     | Contains shortcut fields editable in other sub forms but accessible in first tab easily | N/A |
+| **TypeaheadProductPackCollectionType** | `pack_items` | List of products (for Pack of product) | `AddProductToPackCommand` `UpdateProductPackCommand` `RemoveProductFromPachCommand` |
+| **ProductCategoriesType** | `product_categories` | Manage product categories | N/A |
+
+### BasicInformationType
+
+| Fields | Field type                       | Description            |
+|:-------|:---------------------------------|:-----------------------|
+| `product_id` | `HiddenType` | The *Product* ID |
```
zuk3975, line 41: whats the idea behind product_id here?
after we save a ghost product for first time, we fill this value up?
shouldn't we redirect to edit page after we save the ghost product instead and have an id from request?

src/content/1.7/development/page-reference/back-office/product/form/_index.md
jolelievre: Right I forgot to remove it, thanks

src/content/1.7/development/page-reference/back-office/product/form/_index.md
jolelievre: You're right the id in the request/url is enough, I dropped this one as well

src/content/1.7/development/page-reference/back-office/product/form/_index.md
jolelievre: Done

eternoendless (approved):

eternoendless:
Thanks @jolelievre

src/content/1.7/development/components/form/product/_index.md
```diff
@@ -0,0 +1,239 @@
...
+## Page form
+
+This form combines different sub form types, each one is managed by a dedicated CQRS command (sometimes more).
+
+| Form type                | Field      | Description                                              | CQRS Command                                           |
+|:-------------------------|:-----------|:---------------------------------------------------------|:-------------------------------------------------------|
+| **BasicInformationType** | `basic`    | Contains the basic product information of the Product | `AddProductCommand` and `UpdateBasicProductCommand` |
+| **RelationshipsType**    | `relationships` | Contains relationships of the product | N/A |
+| **ShortcutType**         | `shortcuts`     | Contains shortcut fields editable in other sub forms but accessible in first tab easily | N/A |
+| **TypeaheadProductPackCollectionType** | `pack_items` | List of products (for Pack of product) | `AddProductToPackCommand` `UpdateProductPackCommand` `RemoveProductFromPachCommand` |
```
zuk3975: > RemoveProductFromPachCommand

RemoveProductFrom**Pack**Command
______
Anyway, sorry for bringing this up again, but we need to reconsider whether we create separate add/remove commands or keep it all in one updateProductPack (or something in the middle- like having cleanPackCommand and UpdatePackCommand). Its important to consider which way would be more beneficial for our presentation structure (if we want to refresh the page after updates/deletions or we just handle everything in javascript and pass packed products list for a single command which handles all cases)

Until now we've decided that we should at least create separate command for cleaning the pack.
So the UpdateProductPack will not allow empty array. But if we provide a list without one item, then command will act as RemoveProductFromPack, so the additional command for that might redundant?). Furthermore, we don't need AddProductToPackCommand for same reason.

So the question arises, maybe we should keep only UpdateProductPackCommand (as you've suggested at first I think) which handles all cases depending on packed items array that is passed? (if its empty array we clean the pack, else we insert new pack items instead of old ones)

PR #548, opened 2020-05-27
Add doc about modules & external services

Quetzacoalt91:


src/content/1.7/modules/creation/external-services.md
```diff
@@ -0,0 +1,83 @@
+---
+title: Interacting with APIs
+---
+
+Interacting with APIs
+=====================
```
PierreRambaud, line 6:
```suggestion
# Interacting with APIs
```

src/content/1.7/modules/creation/external-services.md
```diff
@@ -0,0 +1,83 @@
...
+
+There is different options to transfer data between your shop and any external service.
```
Progi1984, line 8:
```suggestion
There are different options to transfer data between your shop and any external service.
```

src/content/1.7/modules/creation/external-services.md
```diff
@@ -0,0 +1,83 @@
...
+
+This will add the resource `reassurance` available into the permissions list, based on the key.
+
+### Module controllers 
+
+As the webservice is only an interface to get and update objects on the database, it does not allow to run complex actions.
+Module controller may be implemented to allow any external service to reach your shop, then trigger specific actions or retrieve content.
+
+You should implement a method that filters non-authenticated calls. This prevents guests accessing private content, or trigger actions on your behalf.
+This can be done by generating your own token and checking it eveytime the controller is called. `Tools::encrypt($passwd)` may be useful.
```
Progi1984, line 62:
```suggestion
This can be done by generating your own token and checking it eveytime the controller is called. `Tools::encrypt($token)` may be useful.
```

src/content/1.7/modules/creation/external-services.md
Quetzacoalt91: This should be done on the whole project, as I used another page as a template.

Quetzacoalt91:
Thanks for the suggestions that I just commited

Progi1984:
Thanks @Quetzacoalt91

PR #549, opened 2020-05-28
Update Order page doc

MatShir:
Add the order preview hook


src/content/1.7/modules/sample-modules/order-view-page-new-hooks/_index.md
```diff
@@ -2,15 +2,15 @@
 title: Order view page new hooks demo tutorial 
 weight: 2
 aliases:
-    - /1.7/modules/sample_modules/order_view_page_new_hooks/
+    - /1.7/modules/sample_modules/order_pages_new_hooks/
```
jolelievre, line 5: Alias are not supposed to be modified, they allow redirection for old page names to avoid 404 error on old links

The weird thing here is that the alias matches the page name so I don't understand its interest..

jolelievre: Ok I get it it's because it used `_` but we need to use `-`

So @MatShir if you wan to change this page url you need to:
- rename this folder into `order-pages-new-hooks`
- leave the current alias as is
- add a new alias:

```
- /1.7/modules/sample_modules/order-view-page-new-hooks/
```

jolelievre: And one last question, it's a good idea to add this hook for preview in listing, but are you sure it's done in the demo module? I thought it only added stuff in the view page, but it would be a good idea to add this in the sample module

Also I think you should add this hook in this page https://devdocs.prestashop.com/1.7/modules/core-updates/1.7.7/#new-hooks along with other new order hooks

MatShir: Yes, it definitely needs to be added in this page as well !
Because the demo hooks are not plugged to all the hooks, I thought it will cool to add the preview so the dev can see all the possibilities around the order pages. 
If you think it isn't relevant, I can remove it.

Progi1984:
Ping @MatShir

MatShir:
"Better late than never"

src/content/1.7/modules/sample-modules/order-pages-new-hooks/_index.md
```diff
@@ -48,10 +49,15 @@ On module installation the following hooks can be registered:
  - `displayAdminOrder` - displayed at the bottom of the Order page
  - `displayAdminOrderTop` - displayed at the top of the Order page
  - `actionGetAdminOrderButtons` - to display additional action buttons into the main buttons bar.
+
+on the listing page:
```
Progi1984, line 53:
```suggestion
And on the listing page:
```

Progi1984 (commented):
A small feedback

Progi1984:
@MatShir Could you rebase your PR ?

kpodemski:
ping @MatShir 

could you resolve conflicts? It would be great to see your changes in docs

src/content/1.7/modules/sample-modules/order-view-page-new-hooks/_index.md
kpodemski: @jolelievre 

https://github.com/PrestaShop/example-modules/pull/31

MatShir:
I wish I could but I can't/don't know how resolve conflict in command line

kpodemski:
@MatShir 

here you go, I did this for you

let's wait for a quick re-review from @jolelievre or @Progi1984

MatShir:
Thanks @kpodemski I appreciate 🤗

Progi1984 (approved):

PR #550, opened 2020-05-28
Add FAQ about overriding the product cover

jolelievre:
In 177 the cover image is used by default in product listing, this explains how to verride this behaviour starting .7.7

### DIFF
--- a/src/content/1.7/faq/product.md
+++ b/src/content/1.7/faq/product.md
@@ -11,8 +11,8 @@ title: Product FAQ
 
 **Q:** How can I override the cover image of my products?
 
-**A:** By default, when a product is displayed in a list its cover image is used, it is configurable in the BackOffice and is set in product properties via the `cover_image_id` property.
-If you want to change this default behaviour you can use the `actionGetProductPropertiesAfter` in your module and change this property.
+**A:** By default, when a product is displayed in a list its cover image is used, it is configurable in the BackOffice and is set in product properties via the `cover_image_id` key.
+If you want to change this default behaviour you can use the `actionGetProductPropertiesAfter` in your module and change this key.
 
 ```php
 /**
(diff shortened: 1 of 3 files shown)


Matt75 (approved):

src/content/1.7/faq/product.md
```diff
@@ -0,0 +1,37 @@
...
+# Product FAQ
+
+- [Product cover](#product-cover)
+
+## Product cover
+{{< minver v="1.7.7" title="true" >}}
+
+**Q:** How can I override the cover image of my products?
+
+**A:** By default, when a product is displayed in a list its cover image is used, it is configurable in the BackOffice and is set in product properties via the `cover_image_id` property.
```
PierreRambaud, line 14: It's not really a property :thinking: 
`$product` is an array, not an object

Progi1984:
```suggestion
**A:** By default, when a product is displayed in a list its cover image is used, it is configurable in the BackOffice and is set in product properties via the `cover_image_id` key.
```

Progi1984 (commented):
Ping @jolelievre

src/content/1.7/faq/product.md
jolelievre: I used the term `property` because it was in the context of the `getProductProperties` function, besides `$product` is an array at the beginning of the workflow and is then converted to a `ProductLazyArray` which is a mix between an array and an object 😅

But I see your point and I agree so to avoid any misunderstanding I used the term `key` instead as suggested

Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
We should wait for the original pull request to be merged before merging this one :)

jolelievre:
Pending core PR https://github.com/PrestaShop/PrestaShop/pull/19437

PR #555, opened 2020-06-02
Adds warning about module console command

Matt75:
Adds warning about console command from module with usage of legacy classes.
Remove example for PrestaShop < 1.7.3 because it's available only from PrestaShop 1.7.5 for modules.

### DIFF
--- a/src/content/1.7/modules/concepts/commands.md
+++ b/src/content/1.7/modules/concepts/commands.md
@@ -15,7 +15,8 @@ Since version 1.7 of PrestaShop, everyone have access to the PrestaShop console 
 Since v1.7.5, you can add and provide your own commands into the PrestaShop console using modules.
 
 {{% notice warning %}}
-If you use PrestaShop Core legacy classes, it will not works yet. We will fix it soon.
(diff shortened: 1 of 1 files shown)


src/content/1.7/modules/concepts/commands.md
```diff
@@ -9,11 +9,15 @@ weight: 10
...
 
 ``
-./bin/console # or ./app/console for PrestaShop < 1.7.3
+./bin/console
 ``
 
 Since v1.7.5, you can add and provide your own commands into the PrestaShop console using modules.
 
+{{% notice warning %}}
+If you use PrestaShop Core legacy classes, it will not works yet. We will fix it soon.
```
matks, line 18: The idea is there but it's a bit "scary" when said like this 😛 may I suggest something ?

```suggestion
If you load and use PrestaShop Core legacy classes such as an ObjectModel within a Command context, you might run into issues. This is a known limitation of the Commands.
Removing this limitation is being explored for future PS versions.
```

matks:
Thank you @Matt75

PR #565, opened 2020-06-10
Document smarty extensions

eternoendless:
I documented most extensions, but a few are yet to be added. My eyeballs were bleeding too much.

See:
- https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/config/smarty.config.inc.php
- https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/config/smartyadmin.config.inc.php
- https://github.com/PrestaShop/PrestaShop/blob/1.7.7.x/config/smartyfront.config.inc.php


PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Thanks @eternoendless

PR #567, opened 2020-06-11
Use actionFrontControllerSetMedia if you want to add js and css

PierreRambaud:
Do not use `displayHeader` to define Js and Css we already have hooks for Login page, front and back office.


matks:
Need @Quetzacoalt91 review

Quetzacoalt91:
Thank you @PierreRambaud

PR #575, opened 2020-06-16
Add children list in themes / distribution section

ksaandev:
https://devdocs.prestashop.com/1.7/themes/distribution/

Looks empty right 😄


PierreRambaud:
Thanks @ksaandev

PR #576, opened 2020-06-16
Add title and children list in theme / reference section

ksaandev:
https://devdocs.prestashop.com/1.7/themes/reference/

Looks empty right 😄 ? Sad ... let's change that !


PierreRambaud:
Thanks @ksaandev

PR #577, opened 2020-06-16
Update Hugo to latest (0.72)

eternoendless:
Hugo [switched its markdown renderer to Goldmark](https://gohugo.io/news/0.60.0-relnotes/) (which is commonmark compliant) in 0.60, so there may be some rendering issues here and there.

Hugo also uses Chroma instead of Pygments for syntax highlighting now. This is good news, because I never got to make native syntax highlighting before and we had to rely on client-side highlighting, which was obviously not optimal. The downside to this change is that Chroma will refuse to highlight PHP code if it doesn't start with `<?php` and they [refuse to fix that](https://github.com/alecthomas/chroma/issues/210), so I had to add that block to all snippets.

This upgrade should allow us to progressively adopt new features as well as benefit from increased build performance.


PierreRambaud:
Thanks @eternoendless

PR #582, opened 2020-06-18
Move webservice into a section of its own

eternoendless:
I think it makes sense to put it at the same level as Modules and Themes

Screenshot 2020-06-18 at 12 44 27


NeOMakinG (approved):
Agree, this is way better

matks:
It makes sense indeed. Thank you @eternoendless

matks (approved):

PR #583, opened 2020-06-18
Rename "Core Development" to "Core Development Reference"

eternoendless:
Makes more sense to me this way because it's not only meant for core development


NeOMakinG (approved):

matks (approved):

matks:
Thank you @eternoendless

PR #586, opened 2020-06-18
updateAddressForm is incorrect

muyncky:
Only `updatedAddressForm` works.

This event is triggered after it is update**d**.


Progi1984:
Thanks @muyncky

PR #588, opened 2020-06-19
Add {{% children %}} in Themes > Reference > Templates

matks:
Following examples like PrestaShop#581 from @ksaandev 😄

Page without children today: https://devdocs.prestashop.com/1.7/themes/reference/templates/


PierreRambaud:
Thanks @matks

PR #593, opened 2020-06-21
Update create user SQL query

shef-er:
IDENTIFIED BY cannot be used in GRANT request

### DIFF
--- a/src/content/1.7/basics/installation/_index.md
+++ b/src/content/1.7/basics/installation/_index.md
@@ -69,7 +69,8 @@ Create the database and give it a name like "prestashop":
 Grant privileges to that database to a new user (the one that PrestaShop will use to connect to the database). Let's call it "prestashopuser".
 
 ```bash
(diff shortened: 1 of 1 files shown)


matks:
Maybe it's some specific database engines or versions ? Because I can see plenty of examples on google using IDENTIFIED with GRANT BY
https://mediatemple.net/community/products/dv/204404494/how-do-i-grant-privileges-in-mysql

Anyway I like the idea of splitting the queries as it makes the example clearer for people (each step does one thing, no mix)

shef-er:
Ok, maybe there is some databases that allows to do such request. I ran the original line in the mysql 8.0.20 and had recived  syntax error

shef-er:
Also there is no IDENTIFIED BY with GRANT request in the official documantation https://dev.mysql.com/doc/refman/8.0/en/grant.html for MySQL

PierreRambaud:
Thanks @shef-er

PR #594, opened 2020-06-22
Add note about actionFrontControllerSetMedia hook after cache section

marekjedrzejewski:
I lost some time, because I skipped over registering this hook.
I think this might help people like me.

### DIFF
--- a/src/content/1.7/modules/creation/displaying-content-in-front-office.md
+++ b/src/content/1.7/modules/creation/displaying-content-in-front-office.md
@@ -263,6 +263,9 @@ site, as it severely slows everything down! You should always perform
 all your tests in a test site, ideally on your own computer rather than
 online.
 
+If you still can't see the changes you made, you might have forgotten to 
+register `actionFrontControllerSetMedia` hook.
+
 Embedding a template in the theme
 ---------------------------------
 


kpodemski:
but it's not necessary to register this hook if you don't have any assets

marekjedrzejewski:
I didn't see styling or css link in head until I went through 'transplant a module' -> mymodule -> transplant to: actionFrontControllerSetMedia.

Seems like I skipped this section:

> The actionFrontControllerSetMedia hook is not part of the visual header, but enables us to put assets after code in the <head> tag of the generated HTML file. This is very useful for JavaScript or CSS files. To add a link to our CSS file in the page’s <head> tag, we use the registerStylesheet() method, which generates the correct <link> tag to the CSS file indicated in parameters. To add our JS script in the page, we use the registerJavascript() method, which generates the correct <script> tag.

And while glancing over the previous steps I also didn't catch it. I don't have any experience with presta, so I may not understand how things should work and why would they work that way. I am simply going through documentation here, and got stuck on this bit, so I thought this might help others like me.

If this was unnecessary step, but worked, what was needed to make it work then?

kpodemski:
@marekjedrzejewski well, i think that this article:
https://devdocs.prestashop.com/1.7/modules/creation/displaying-content-in-front-office/

has all you need, others just can't skip beginning of the article and they should be ok :)

marekjedrzejewski:
That's the one I'm trying to add note to...
I was thinking about something similar to notes in [django tutorial](https://docs.djangoproject.com/en/3.0/intro/tutorial01/), which has things like: 

> Page not found?
> 
> If you get an error page here, check that you’re going to http://localhost:8000/polls/ and not http://localhost:8000/.

Things that are not essential. In this case the url is even in the line above, but they can help if someone gets stuck.

matks:
Thank you @marekjedrzejewski

PR #598, opened 2020-06-23
Correct path to load Module twig

camlafit:
As explained in https://www.prestashop.com/forums/topic/1017862-using-twig-tmplate-from-a-module-hook-in-176/?tab=comments#comment-3210727

Twig template are their root path

| Questions     | Answers
| Description?  | Documentation is wrong, generate a 500 
| Type?         | improvement
| BC breaks?    | no
| Deprecations? | no
| Fixed ticket? | Fixes {paste the issue here}.
| How to test?  | Follow actual documentation about twig template load from any module.


PierreRambaud:
Thanks @camlafit

PR #600, opened 2020-06-24
Provide example of breaking changes about jQuery update

Quetzacoalt91:
| Questions     | Answers
| Description?  | Even though we use jQuery Migrate on PS 1.7.7, modules need to be checked and updated to prevent compatibility issues. This PR provides example of breaking changes.
| Type?         | improvement
| BC breaks?    | Yes
| Deprecations? | No
| Fixed ticket? | /
| How to test?  | /


matks:
Thank you @Quetzacoalt91

PR #628, opened 2020-07-10
Improve style of quote blocks

eternoendless:
| Questions     | Answers
| Description?  | Improved style for quote blocks + fixed all wrong use of quote blocks where it was actually a notice.
| Fixed ticket? | ~

Before:
Screenshot 2020-07-10 at 17 57 47

After:
Screenshot 2020-07-10 at 17 58 04


src/themes/hugo-theme-learn/_src/sass/components/_blockquote.scss
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  ul, ol {
+    margin: {
+      top: .5rem;
+      bottom: .5rem;
+    }
+
+    li {
+      margin-bottom: 0 !important;
```
matks, line 27: I'm very bad at html/css but I was always told never to use `!important` 😬 😄

matks (approved):
Code seems good but needs to be tested. We recently had issues with @NeOMakinG about pushing assets to be acknowledged by GitHub Pages.

src/themes/hugo-theme-learn/_src/sass/components/_blockquote.scss
NeOMakinG: Well, did you try to select it a bit more precisely or using the selector causing you to be forced to use an important ?

NeOMakinG:
> Code seems good but needs to be tested. We recently had issues with @NeOMakinG about pushing assets to be acknowledged by GitHub Pages.

Pushing CSS is working fine, JS was not, but JS is now builded inside travis before serving ! No more problems about this

Also, we could kick assets from this repository as I think CSS assets are also builded by travis

NeOMakinG (approved):

src/themes/hugo-theme-learn/_src/sass/components/_blockquote.scss
eternoendless: I just copied and pasted the original css from "notice". That said, the `important` is probably justified, as there was probably another rule messing with `li` attached to a classnamed parent element, which takes precedence.

The whole theme would clearly benefit form a revamp/

PierreRambaud (approved):

PierreRambaud:
Thanks @eternoendless

PR #645, opened 2020-07-24
It's @Modules, not @Module

elboletaire:
| Questions     | Answers
| Description?  | Tiny error stating that `@Module` should be used to access modules templates, but it's `@Modules` actually
| Fixed ticket? | no ticket afaik


Progi1984:
Thanks @elboletaire

PR #654, opened 2020-08-03
Fix wrong compatibility note

eternoendless:
| Questions     | Answers
| Description?  | Fix wrong compatibility note. `getAdminLink()` does not support parameters in 1.6, so we shouldn't recommend that.
| Fixed ticket? | ~


matks:
Thank you @eternoendless

Issue #653, opened 2020-08-03
Add search in the hooks page


eternoendless:
The hook list is very long, how about adding a search box so that people can quickly filter the list by searching the ones that a certain string?

Progi1984:
Done ;)

eternoendless:
❤️ ❤️ ❤️

PR #656, opened 2020-08-05
Add search hooks

NeOMakinG:
| Questions     | Answers
| Description?  | We wants to be able to search inside the hooks page in order to facilitate the use of this page
| Test?  | Someone should test before merging please! This is a new feature and I'm not sure about wording and design
| Fixes? | https://github.com/PrestaShop/docs/issues/653

image

image


src/themes/hugo-theme-learn/static/js/hookFilter.js
```diff
@@ -0,0 +1,53 @@
...
+      if (matchesSearch) {
+        count += 1;
+      }
+    });
+
+    if (count === 0) {
+      $('#hookFilter .empty').addClass('show');
+    } else {
+      $('#hookFilter .empty').removeClass('show');
+    }
```
Progi1984, line 39:
```suggestion
    $('#hookFilter .empty').toggleClass('show', count === 0);
```

matthieu-rolland: In his condition the class `show` is only added if `count === 0`, with `toggleclass` it could remove the class instead, if it was present before, so the behavior is not exactly the same. Is this really the behavior we want ?

Progi1984:
Thanks @NeOMakinG

matks:
Did not have the time to do the QA on the PR 😅  but in production it seems great 😆

PR #677, opened 2020-08-21
Fix incorrect terms usage. Сlarify using `dump()`.

likemusic:
1. Fix incorrect use of terms `function` and `method`. Functions - without classes. Methods - with classes.
2. Сlarify using `dump()` function. In code we call [dump()](https://github.com/symfony/var-dumper/blob/master/Resources/functions/dump.php#L18) function, but not [VarDumper::dump()](https://github.com/symfony/var-dumper/blob/master/VarDumper.php#L37) method.

| Questions     | Answers
| Description?  |1. Fix incorrect use of terms `function` and `method`. Functions - without classes. Methods - with classes. <br/>2. Сlarify using `dump()` function. In code we call [dump()](https://github.com/symfony/var-dumper/blob/master/Resources/functions/dump.php#L18) function, but not [VarDumper::dump()](https://github.com/symfony/var-dumper/blob/master/VarDumper.php#L37) method.
| Fixed ticket? | No

### DIFF
--- a/src/content/1.7/basics/installation/configuration.md
+++ b/src/content/1.7/basics/installation/configuration.md
@@ -32,15 +32,15 @@ You, on the other hand, need this information in order to correct any potential 
 define('_PS_MODE_DEV_', true);
 ```
 
-## Using the dump() method
+## Using the dump() function
 
-These two methods enable you to check for the state of a given variable at a specific place within your code.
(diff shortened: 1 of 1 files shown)


PierreRambaud (approved):

NeOMakinG:
Thanks @likemusic

PR #682, opened 2020-08-24
Fix add() definition according to source phpDoc.

likemusic:
| Questions     | Answers
| Description?  | Fix add() definition according to [source phpDoc](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/ObjectModel.php#L521)
| Fixed ticket? | No


NeOMakinG (approved):

NeOMakinG:
Thanks @likemusic

PR #690, opened 2020-08-28
Allow installation of Hugo's mac os version via install script

matthieu-rolland:
| Questions     | Answers
| Description?  | On mac os, hugo should be installed via the installation script instead of brew, so that everyone use the same version.
| Fixed ticket? | none

### DIFF
--- a/bin/installHugo.sh
+++ b/bin/installHugo.sh
@@ -3,15 +3,13 @@ set -euo pipefail
 
 export HUGO_VERSION="0.72.0"
 
-if [ `uname` = "Linux" ]
-then
-	export HUGO_OS="Linux-64bit"
-elif [ `uname` = "Darwin" ]
-then
+if [ `uname` = "Linux" ]; then
+    export HUGO_OS="Linux-64bit"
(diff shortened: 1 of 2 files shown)


bin/installHugo.sh
```diff
@@ -3,9 +3,21 @@ set -euo pipefail
...
+if [ `uname` = "Linux" ]
+then
+	export HUGO_OS="Linux-64bit"
+elif [ `uname` = "Darwin" ]
+then
+    export HUGO_OS="macOS-64bit"
+else
+	echo 'This script only works on Linux and Mac os'
+	exit 1
+fi
```
PierreRambaud, line 15:
```suggestion
if [ `uname` = "Linux" ]; then
    export HUGO_OS="Linux-64bit"
elif [ `uname` = "Darwin" ]; then
    export HUGO_OS="macOS-64bit"
else
    echo 'This script only works on Linux and Mac os'
    exit 1
fi
```

PierreRambaud (approved):

PierreRambaud:
Thanks @matthieu-rolland

PR #695, opened 2020-08-30
Svg images instead of png.

likemusic:
| Questions     | Answers
| Description?  | Create and use svg-images instead of png for schemas drawn in draw.io (because svg-images are scaled without loss of quality).
| Fixed ticket? | No.

By the way I have made some small fixes in xml-files and split form-schema to 2 separate schemas (for display and submit).


NeOMakinG (approved):
Should be fine for me, thanks :)

Issue #342, opened 2019-09-13
Improve Webservice documentation


eternoendless:
- Document all or the main models
- Show examples (sync product catalogue)

eternoendless:
Related: Missing variable in documentation about multistore and webservice #297

eternoendless:
Use tutorials from 1.6: http://doc.prestashop.com/display/PS16/Web+service+tutorial (there are 10 chapters + cheat sheet)

matks:
@jolelievre is this topic finished? 😊

PR #726, opened 2020-09-14
Introducing new webservice documentation resources

PierreRambaud:
| Questions     | Answers
| Description?  | Initial commit for webservices resrouces


src/content/1.7/webservice/resources/content_management_system.md
```diff
@@ -0,0 +1,43 @@
...
+| **id_cms_category**  | isUnsignedInt |          |           |               |                |             |
+| **position**         |               |          |           |               |                |             |
+| **indexation**       |               |          |           |               |                |             |
+| **active**           |               |          |           |               |                |             |
+| **meta_description** | isGenericName |          |           | 512           |                |             |
+| **meta_keywords**    | isGenericName |          |           | 255           |                |             |
+| **meta_title**       | isGenericName | true     |           | 255           |                |             |
+| **head_seo_title**   | isGenericName |          |           | 255           |                |             |
+| **link_rewrite**     | isLinkRewrite | true     |           | 128           |                |             |
+| **content**          | isCleanHtml   |          |           | 3999999999999 |                |             |
```
Progi1984, line 20:
```suggestion
| **content**          | isCleanHtml   |          |           | 3999999999999 |                |             |
```

3999999999999 : Really ?

PierreRambaud: It's the value in the webservice API and in the code :sweat_smile:

PierreRambaud:
```
'content' => ['type' => self::TYPE_HTML, 'lang' => true, 'validate' => 'isCleanHtml', 'size' => 3999999999999],
```

src/content/1.7/webservice/resources/images.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Images
+---
+
+# Resources for Images
```
Progi1984, line 5: Empty file ?

src/content/1.7/webservice/resources/addresses.md
```diff
@@ -0,0 +1,67 @@
+---
+title: Addresses
+---
+
+# Resources for Addresses
+
+### Address
+
+|        Name         |       Format       | Required | Read Only | Max size | Not filterable | Description |
```
jolelievre, line 9: Just a remark: about read only it appears most field are writable and only a handful of them are read only Besides I think people who read this documentation are more interested about things they can modify rather than those they can't So should we display a `writable` column instead of a `read-only` one

Also many rows end up with empty values, maybe it would be easier to read if we used some icons instead of `true` ? Like ✅   and ❌   (or similar) and populate each cells

I have the same questions regarding `Not filterable` column, especially when all the fields of an API appear to be filterable. Maybe we should only display this column when some fields have special values (same for read only). And add a note a the end of the table when it's not displayed like:
`All fields are filterable`
`All fields are writable`

This would make the table smaller and easier to read

src/content/1.7/webservice/resources/addresses.md
```diff
@@ -0,0 +1,67 @@
+---
+title: Addresses
+---
+
+# Resources for Addresses
```
jolelievre, line 5: Would it be possible to add an example of the listing for each resource as well? Only the specified resource is shown for now And I believe some listings have special display (like associations and stuffs) that are worth documenting

src/content/1.7/webservice/resources/carts.md
```diff
@@ -0,0 +1,71 @@
...
+	<recyclable><![CDATA[]]></recyclable>
+	<gift><![CDATA[]]></gift>
+	<gift_message><![CDATA[]]></gift_message>
+	<mobile_theme><![CDATA[]]></mobile_theme>
+	<delivery_option><![CDATA[]]></delivery_option>
+	<secure_key><![CDATA[]]></secure_key>
+	<allow_seperated_package><![CDATA[]]></allow_seperated_package>
+	<date_add><![CDATA[]]></date_add>
+	<date_upd><![CDATA[]]></date_upd>
+<associations>
```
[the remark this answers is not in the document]
jolelievre: From what I see in the code the infos could be retrieved from the class themselves ib some cases:

```php
    'associations' => [
            'cart_rows' => [
                'resource' => 'cart_row',
                'virtual_entity' => true,
                'fields' => [
                    'id_product' => ['required' => true, 'xlink_resource' => 'products'],
                    'id_product_attribute' => ['required' => true, 'xlink_resource' => 'combinations'],
                    'id_address_delivery' => ['required' => true, 'xlink_resource' => 'addresses'],
                    'id_customization' => ['required' => false, 'xlink_resource' => 'customizations'],
                    'quantity' => ['required' => true],
                ],
            ],
        ],
```

But sometimes it relies on special getters:

```php
'associations' => [
            'categories' => ['getter' => 'getChildrenWs', 'resource' => 'category'],
            'products' => ['getter' => 'getProductsWs', 'resource' => 'product'],
        ],
```

So I guess we could make a script to document those that only rely on configurations, and manually deal with the ones that rely on getters. But anyway this call all be done in another PR.

src/content/1.7/webservice/resources/customizations.md
```diff
@@ -0,0 +1,54 @@
+---
+title: Customizations
+---
+
+# Resources for Customizations
+
+### Customization
+
+|           Name           |    Format    | Required | Read Only | Max size | Not filterable | Description |
```
jolelievre, line 9: For this kind of API the `Max size` column might be useless as well

src/content/1.7/webservice/resources/addresses.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Address
+
+|        Name         |       Format       | Required | Read Only | Max size | Not filterable | Description |
+| :------------------ | :----------------- | :------- | :-------- | :------- | :------------- | :---------- |
+| **id_customer**     | isNullOrUnsignedId |          |           |          |                |             |
+| **id_manufacturer** | isNullOrUnsignedId |          |           |          |                |             |
```
jolelievre, line 12: I think we might be able to auto generate descriptions for all the ID fields based on their name

`id_manufacturer` => `Manufacturer ID`

Of course they may need additional info, in this address example we might need to add details like `Only use for address related to a manufacturer` But most cases would already be enough I guess

jolelievre: I also wonder if we should add a link to the API matching the entity type or is it overkill?

src/content/1.7/webservice/resources/images.md
PierreRambaud: This file doesn't have synopsis or blank parameters :(

matks:
This PR somehow solves part of https://github.com/PrestaShop/docs/issues/342 isn't it ?

matks:
By the way **THIS IS AWESOME** ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

src/content/1.7/webservice/resources/addresses.md
PierreRambaud: I change it with emojis as you requested, it's a good idea.
But for "All fields are filterable / writable", if we add a not writable field, we will have to change the whole doc instead of changing one line :sweat_smile:

src/content/1.7/webservice/resources/carts.md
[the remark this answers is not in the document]
PierreRambaud: Should be in another PR since the API returns nothing about associations :sweat_smile:

src/content/1.7/webservice/resources/addresses.md
PierreRambaud: What do you mean? Something like that 
image ?

src/content/1.7/webservice/resources/customizations.md
PierreRambaud: Maybe, but it's a generated file.
I will try to improve this depending on the values we have in the results.

src/content/1.7/webservice/resources/addresses.md
PierreRambaud: > id_manufacturer => Manufacturer ID

I will do it :+1: 

I think the link is overkill right now ^^

src/content/1.7/webservice/resources/search.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Search
+---
+
+# Resources for Search
```
Progi1984, line 5: Empty page

src/content/1.7/webservice/resources/customizations.md
PierreRambaud: I remove all unneeded columns when it's possible.

src/content/1.7/webservice/resources/addresses.md
jolelievre: Yep unless they all look very basic and you think they are not very useful?

PierreRambaud: I think it's not really useful :thinking:

jolelievre (approved):

Progi1984 (approved):

Progi1984:
Good job @PierreRambaud

PR #750, opened 2020-09-30
Explain how to use Router component

prestaforum:
| Questions     | Answers
| Description?  | I need to use the JS router and found no documentation about it. A nice person on IRC helped me use it but it's better to add it in doc
| Fixed ticket? |


NeOMakinG (approved):
I think we should also add the router inside the new prestashop.components object too ?

wdyt @PrestaShop/prestashop-maintainers ?

PierreRambaud:
> I think we should also add the router inside the new prestashop.components object too ?
> 
> wdyt @PrestaShop/prestashop-maintainers ?

You're right

NeOMakinG:
Thanks @prestaforum

PR #757, opened 2020-10-03
Updating link for active record pattern

ThanasisMpalatsoukas:
The link was referencing a wrong page from wikipedia.

| Questions     | Answers
| Description?  | Changes on the documentation
| Fixed ticket? | Fixed a mistake on an link which pointed to an non-existing article

### DIFF
--- a/src/content/1.7/development/database/objectmodel.md
+++ b/src/content/1.7/development/database/objectmodel.md
@@ -7,7 +7,7 @@ weight: 3
 
 When needing to dive deep, you have to use the ObjectModel class. This is the main object of PrestaShop’s object model. It can be overridden… with precaution.
 
-It is an Active Record kind of class (see: https://en.wikipedia.org/wiki/Active_record_pattern). The table attributes or view attributes of PrestaShop’s database are encapsulated in this class. Therefore, the class is tied to a database record. After the object has been instantiated, a new record is added to the database. Each object retrieves its data from the database; when an object is updated, the record to which it is tied is updated as well. The class implements accessors for each attribute.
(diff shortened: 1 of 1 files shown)


src/content/1.7/development/database/objectmodel.md
```diff
@@ -7,7 +7,7 @@ weight: 3
 
 When needing to dive deep, you have to use the ObjectModel class. This is the main object of PrestaShop’s object model. It can be overridden… with precaution.
 
-It is an Active Record kind of class (see: https://en.wikipedia.org/wiki/Active_record_pattern). The table attributes or view attributes of PrestaShop’s database are encapsulated in this class. Therefore, the class is tied to a database record. After the object has been instantiated, a new record is added to the database. Each object retrieves its data from the database; when an object is updated, the record to which it is tied is updated as well. The class implements accessors for each attribute.
+It is an Active Record kind of class (see: https://en.wikipedia.org/wiki/Active_record_pattern ). The table attributes or view attributes of PrestaShop’s database are encapsulated in this class. Therefore, the class is tied to a database record. After the object has been instantiated, a new record is added to the database. Each object retrieves its data from the database; when an object is updated, the record to which it is tied is updated as well. The class implements accessors for each attribute.
```
matks, line 10:
```suggestion
It is an Active Record kind of class (see: [Active record pattern](https://en.wikipedia.org/wiki/Active_record_pattern)). The table attributes or view attributes of PrestaShop’s database are encapsulated in this class. Therefore, the class is tied to a database record. After the object has been instantiated, a new record is added to the database. Each object retrieves its data from the database; when an object is updated, the record to which it is tied is updated as well. The class implements accessors for each attribute.
```

matks (approved):
Hi ! nice catch ! I converted this hard-coded link into a Markdown link

matks:
Thank you @ThanasisMpalatsoukas

PR #761, opened 2020-10-04
Add Known Limit about images url rewriting

okom3pom:
| Questions     | Answers
| Description?  | Add Known limit for images
| Fixed ticket? | Relative to PrestaShop/PrestaShop#21196

### DIFF
--- a/src/content/1.7/basics/installation/system-requirements.md
+++ b/src/content/1.7/basics/installation/system-requirements.md
@@ -220,3 +220,29 @@ extension = php_gd2.dll
 allow_url_fopen = On
 allow_url_include = Off
 ```
+
+### Known limits
+
+With url rewrite enable, you can't use more than 9 999 999 images.
+
+You can exceed this limit by adding 
+
+```ini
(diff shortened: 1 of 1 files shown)


okom3pom:
https://github.com/PrestaShop/PrestaShop/pull/21196

okom3pom:
I close it ? @Progi1984  && @matks ?

matks:
Hi ! I dont think this should go in "system requirements" because this is not a system requirement. People browsing this page are looking for "what do I need to run PrestaShop", not known limitations of the software for really big catalogs.

This looks like an accurate answer to an accurate question so I suggest to move it into the [FAQ section](https://devdocs.prestashop.com/1.7/faq/).

It could be for example:

**Question**

Why can I not use more than 9 999 999 images?

**Answer**

The limitation of the `.htaccess` file prevent the use of more than 9 999 999 images because ...

You can remove this limitation by adding the needed rules to go further. Example ...

okom3pom:
I will move in the FAQ

okom3pom:
https://github.com/PrestaShop/docs/pull/928

PR #766, opened 2020-10-06
Valid path to unit tests

kpodemski:
| Questions     | Answers
| Description?  | There's no such a file as check_phpunit.sh
| Fixed ticket? | n/a
| Sponsor company | impSolutions

### DIFF
--- a/src/content/1.7/contribute/contribute-pull-requests/contribute_using_docker.md
+++ b/src/content/1.7/contribute/contribute-pull-requests/contribute_using_docker.md
@@ -103,7 +103,7 @@ Before all, you may ensure your changes don't break everything: this is why we h
 You can execute it in your dockerized PrestaShop application without altering your website (it uses a specific database).
 
 ```
-docker exec prestashop-git sh tests/check_phpunit.sh
+docker exec prestashop-git sh tests/check_unit.sh
 ```
 
 ### Publish your work
--- a/src/content/1.7/contribute/contribute-pull-requests/contribute_using_docker.md
+++ b/src/content/1.7/contribute/contribute-pull-requests/contribute_using_docker.md
@@ -98,7 +98,7 @@ A good practice is to write meaningful commits messages: it's better to have "co
 ### Launch the test suite
 
 Your changes now sounds ok, and you're almost ready to share your changes with the community.
-Before all, you may ensure your changes don't break everything: this is why we have multiple test suites you can use. Want to read more about tests in PrestaShop? Head to [this](https://devdocs.prestashop.com/1.7/testing/introduction/) page.
+Before all, you may ensure your changes don't break everything: this is why we have multiple test suites you can use. Want to read more about tests in PrestaShop? Head to [this]({{< ref "1.7/testing/introduction.md" >}} page.
 
 You can execute it in your dockerized PrestaShop application without altering your website (it uses a specific database).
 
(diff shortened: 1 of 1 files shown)


matks:
Clearly the doc is wrong however I think we now have 2 such files 😬 
https://github.com/PrestaShop/PrestaShop/blob/develop/tests-legacy/check_phpunit.sh

kpodemski:
@matks oh, right, I forgot about legacy tests 🤦‍♂️ what's your recommendation? I don't think I've been in this part of the docs and I'm trying to follow process as I'm training someone new to PrestaShop :) I have dozen of notes for improvement but this was a quickest one 😅

matks:
I think this part of the documentation was written long ago, when only phpunit was used. We now use PHPUnit and Behat and Playwright
https://devdocs.prestashop.com/1.7/testing/introduction/

As the topic grew bigger I would suggest simply using a link inside this "how to contribute using docker" page. The link would go to Testing page. This way, no matter how the testing chapter evolves, it'll always be accurate.

matks:
@kpodemski so what do we do with this PR ? 😄

kpodemski:
@matks this is example of running `something` inside docker container, it doesn't have much to do with real testing from what I see, so I'd go with this change just to have tests executed without error :)

kpodemski:
@matks 

Or we can add little annotation:

"PS: Want to read more about tests in PrestaShop? Head to this page"

PierreRambaud (approved):

Progi1984 (approved):

kpodemski:
is it OK for you @matks ? 🙂

matks:
> @matks
> 
> Or we can add little annotation:
> 
> "PS: Want to read more about tests in PrestaShop? Head to this page"

Agreed 😊 we can adapt the sentence that is a little bit above:
`Before all, you may ensure your changes don't break everything: this is why we have a test suite you can use.`
⬇️ 
`Before all, you may ensure your changes don't break everything: this is why we have multiple test suites you can use.` (and then your part)

kpodemski:
@matks done, you can merge if it's 👍 for you :)

src/content/1.7/contribute/contribute-pull-requests/contribute_using_docker.md
```diff
@@ -98,12 +98,12 @@ A good practice is to write meaningful commits messages: it's better to have "co
 ### Launch the test suite
 
 Your changes now sounds ok, and you're almost ready to share your changes with the community.
-Before all, you may ensure your changes don't break everything: this is why we have a test suite you can use.
+Before all, you may ensure your changes don't break everything: this is why we have multiple test suites you can use. Want to read more about tests in PrestaShop? Head to [this](https://devdocs.prestashop.com/1.7/testing/introduction/) page.
```
matks, line 101:
```suggestion
Before all, you may ensure your changes don't break everything: this is why we have multiple test suites you can use. Want to read more about tests in PrestaShop? Head to [this]({{< ref "1.7/testing/introduction.md" >}} page.
```

One last change 😉 to avoid hard-coded links. This allows you to run and use the devdocs in localhost

matks (changes requested):

kpodemski:
@matks done, I didn't know about dynamic URLs, thanks 👍

matks (approved):
All good!

PR #782, opened 2020-10-19
Add tinymce config extend documentation

NeOMakinG:
| Questions     | Answers
| Description?  | TinyMCE config is now extendable by external developers, we need to document it

### DIFF
--- a/src/content/1.7/development/components/tinymce.md
+++ b/src/content/1.7/development/components/tinymce.md
@@ -1,0 +1,20 @@
+---
+title: TinyMCE
+menuTitle: TinyMCE
+---
+
+# Modify the TinyMCE configuration
+
+You are able to modify the configuration by using the `actionAdminControllerSetMedia` hook.
+
+Inside the JS file you added with this hook, you need to create a global object containing your custom configuration :
+
+```js
+window.defaultTinyMceConfig = {
+  [...]
+}
+```
+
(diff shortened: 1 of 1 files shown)


src/content/1.7/development/components/tinymce.md
```diff
@@ -0,0 +1,20 @@
+---
+title: TinyMCE
+menuTitle: TinyMCE
+---
+
+# Modify the TinyMCE configuration
+
+You are able to modify the configuration by using the `actionAdminControllerSetMedia` hook.
+
+Inside the JS file you added with this hook, you need to create a global object containing your custom configuration :
```
atomiix, line 10:
```suggestion
Inside the JS file you added with this hook, you need to create a global object containing your custom configuration:
```

src/content/1.7/development/components/tinymce.md
```diff
@@ -0,0 +1,20 @@
...
+
+```js
+window.defaultTinyMceConfig = {
+  [...]
+}
+```
+
+and as you can see in the [tinymce core file](https://github.com/PrestaShop/PrestaShop/blob/0046bf590f033e2ad00594efd92873bc577bf81a/js/admin/tinymce.inc.js) it will load it instead of the default core configuration.
+
+Keep in mind that the config is not completely replaced, it's an `Object.assign`, this means that if you want to remove a certain configuration, you'll have to add it inside you'r own config, overwise it will be kept.
```
atomiix, line 20:
```suggestion
Keep in mind that the config is not completely replaced, it's an `Object.assign`, this means that if you want to remove a certain configuration, you'll have to add it inside your own config, overwise it will be kept.
```

atomiix (approved):

PR #788, opened 2020-10-25
Update nginx.md performance and requirements

gnovaro:
- Update php version to min required
- Performance improvement add http2 for https connections

### DIFF
--- a/src/content/1.7/basics/installation/nginx.md
+++ b/src/content/1.7/basics/installation/nginx.md
@@ -18,7 +18,7 @@ server {
     # listen [::]:80;
 
     # SSL Ipv4 & v6
-    # listen 443 ssl;
+    # listen 443 ssl http2;
     # listen [::]:443 ssl;
 
     # Your SSL Certificates, don't forget to take a look at Certbot (https://certbot.eff.org)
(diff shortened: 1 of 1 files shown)


src/content/1.7/basics/installation/nginx.md
```diff
@@ -18,7 +18,7 @@ server {
     # listen [::]:80;
 
     # SSL Ipv4 & v6
-    # listen 443 ssl;
+    # listen 443 ssl http2;
```
matks, line 21: I'm unfortunately a nginx newbie 🤔  so I dont know if this is valid.

I'm sure it's a good idea 😄  but this file aims to be as generic as possible so it can be used by a very large audience so we cannot put specific items in it.

Help @Quetzacoalt91 @djodjo3 ?

gnovaro: Its a module maybe in some old setup are not installed by default 
https://nginx.org/en/docs/http/ngx_http_v2_module.html
https://http2.akamai.com/demo
https://www.tecmint.com/enable-http-2-in-nginx/

PierreRambaud: It's ok, it's the new standard :)

PierreRambaud (approved):

PierreRambaud:
Thank you @gnovaro

PR #793, opened 2020-10-28
Introduce maintainer processes

matks:
| Questions     | Answers
| Description?  | Introduce maintainer processes with security requirements
| Fixed ticket? |

### DIFF
--- a/src/content/1.7/project/maintainers-guide/processes.md
+++ b/src/content/1.7/project/maintainers-guide/processes.md
@@ -1,0 +1,14 @@
+---
+title: Processes
+---
+
+# Processes
+
+## Security
+
+Maintainer GitHub accounts are granted extended rights on the PrestaShop organization.
+
+We require from maintainers to take the necessary steps to make sure this account is protected. This means, but not only:
+- enable 2FA on GitHub
+- use a safe and strong password
+- be very careful about GitHub tokens granting account accesses
--- a/src/content/1.7/project/maintainers-guide/processes.md
+++ b/src/content/1.7/project/maintainers-guide/processes.md
@@ -4,7 +4,7 @@ title: Processes
 
 # Processes
 
-## Security
+## Security recommendations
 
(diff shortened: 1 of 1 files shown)


kpodemski (approved):

PululuK (dismissed):
@matks are these recommendations or requirements?

If these are the requirements, does that mean that the candidate must meet these points before having his application approved? So how can we verify this?

matks:
> @matks are these recommendations or requirements?

Recommendations for now. It could become requirements in the future.

> If these are the requirements, does that mean that the candidate must meet these points before having his application approved? So how can we verify this?

I dont think we need to be so strict. What is important here is to highlight that maintainer rights on an organization makes the account an important asset not to be underestimated. [Some people go even further](https://twitter.com/TobiasNyholm/status/1315355043185537024) 😅 .

Even if tomorrow these become hard requirements I dont think we need maintainer to meet these points before submitting application and verify it. A ill-intentioned maintainer candidate could enable 2FA then disable it later if he wanted to trick us.

One of the important parts of being part of a maintainer team is trust 😊 . We decided through a vote that you and @kpodemski have earned our trust and could join us as maintainers. If we start verifying that indeed you enabled 2FA even though you said it, it means the trust is gone and the situation is way worse than it looks. We will not verify informations you provide us unless a very nasty situation appears.

PululuK:
> Recommendations for now. It could become requirements in the future.
> 

It does not match your PR description. Could you update it then to avoid ambiguities?

src/content/1.7/project/maintainers-guide/processes.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Processes
+---
+
+# Processes
+
+## Security
```
PululuK, line 7:
```suggestion
## Security recommendations
```

Progi1984 (approved):

Progi1984:
Thanks @matks

PR #804, opened 2020-11-10
Fix PS minimal required version

comxd:
| Questions     | Answers
| Description?  | Prestashop minimal required version fixes
| Fixed ticket? | N/A

### DIFF
--- a/src/content/1.7/modules/concepts/hooks/list-of-hooks.md
+++ b/src/content/1.7/modules/concepts/hooks/list-of-hooks.md
@@ -2480,7 +2480,7 @@ action&lt;HookName>Save
     
 action&lt;GridDefinitionId>GridDefinitionModifier
 : 
-    Available since: {{< minver v="1.7.5" >}}
+    Available since: {{< minver v="1.7.6" >}}
     Located in: /src/Core/Grid/Definition/Factory/AbstractGridDefinitionFactory.php
 
     Parameters:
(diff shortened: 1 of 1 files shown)


src/content/1.7/modules/concepts/hooks/list-of-hooks.md
```diff
@@ -2493,7 +2493,7 @@ action&lt;GridDefinitionId>GridDefinitionModifier
     
 action&lt;GridDefinitionId>GridQueryBuilderModifier
```
matks, line 2494: Hi, these hooks are available since 1.7.5 as far as I know 🤔 
Look at https://github.com/PrestaShop/PrestaShop/blob/1.7.5.x/src/Core/Grid/Data/Factory/DoctrineGridDataFactory.php#L90 for example

matks (changes requested):

src/content/1.7/modules/concepts/hooks/list-of-hooks.md
comxd: Oh! you are right!
I will correct my PR, but I noticed that the hooks were registered in the "hooks" table only since version 1.7.6 as you can see at https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/install-dev/upgrade/sql/1.7.6.0.sql

comxd: After checking the other hooks, this pull request can be canceled because all are available since 1.7.5.

matks: No worries, thanks for the PR it was a good intent 😉

PR #812, opened 2020-11-24
Better title UIKit section

matks:
| Questions     | Answers
| Description?  | Only 1 item has "The" in front of it in "Core reference" category so I suggest to remove it
| Fixed ticket? | 

See https://devdocs.prestashop.com/1.7/development/

Only item "The UIKit" has "The" in front of it. So I suggest to remove it

### DIFF
--- a/src/content/1.7/development/uikit.md
+++ b/src/content/1.7/development/uikit.md
@@ -1,5 +1,5 @@
 ---
-title: The UIKit
+title: UIKit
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):

Progi1984 (approved):

Progi1984:
Thanks @matks

PR #818, opened 2020-11-26
Wrong path on overrading templates

OliverCG:
| Questions     | Answers
| Description?  | In the "Identify the template to override" section, there was an error since the path didn't include the CatalogPage folder which could be very confusing because the other paths in the guide are well formed.
| Fixed ticket? | No, as far as I know.


NeOMakinG (approved):

PululuK (approved):

PululuK:
Thanks @OliverCG

PR #850, opened 2021-01-27
Add warning on registerJavascript / registerStylesheet about potential collisions

Quetzacoalt91:
| Questions     | Answers
| Description?  | I found out that our modules were sometimes non working because of the same IDs used while calling `registerJavascript` or `registerStylesheet`. This PRs explains why a prefix is needed
| Fixed ticket? | /

### DIFF
--- a/src/content/1.7/themes/getting-started/asset-management/_index.md
+++ b/src/content/1.7/themes/getting-started/asset-management/_index.md
@@ -42,7 +42,7 @@ In order to have the most extensible signatures, these 2 methods take 3 argument
 
 This unique identifier needed for each asset. Reusing the same ID is useful to either override or unregister something already loaded by the Core or a native module.
 
-However, avoid generic names when adding new JS and CSS files to avoid colision with other extensions. Prefixing with your module or theme name is a good start.
+However, avoid generic names when adding new JS and CSS files to avoid collision with other extensions. Prefixing with your module or theme name is a good start.
 
 **Relative path**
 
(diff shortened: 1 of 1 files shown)


src/content/1.7/themes/getting-started/asset-management/_index.md
```diff
@@ -40,7 +40,9 @@ In order to have the most extensible signatures, these 2 methods take 3 argument
 
 **ID**
 
-This unique identifier needed for each asset. This is useful to either override or unregister something already loaded by the Core or a native module.
+This unique identifier needed for each asset. Reusing the same ID is useful to either override or unregister something already loaded by the Core or a native module.
+
+However, avoid generic names when adding new JS and CSS files to avoid colision with other extensions. Prefixing with your module or theme name is a good start.
```
matks, line 45:
```suggestion
However, avoid generic names when adding new JS and CSS files to avoid collision with other extensions. Prefixing with your module or theme name is a good start.
```

So we want to tell people to avoid doing something like `$this->registerStylesheet('script');` or `this-> registerStylesheet('jqueryplugin');` ?

Quetzacoalt91: Indeed.

This method has to be called with more than one parameter, but yes, we want generic values to be avoided as several modules could use the same, making the first file added overwritten by the second one.

Quetzacoalt91: Thanks for the suggestion

matks (approved):

matks:
Thank you @Quetzacoalt91

PR #859, opened 2021-02-12
Fix error in executeS return description and add a notice

jf-viguier:
- executeS return an array of associative arrays
- add a notice to use it only for read queries

| Questions     | Answers


Progi1984 (approved):

PierreRambaud (approved):

PierreRambaud:
Thank you @jf-viguier

PR #869, opened 2021-02-20
Little improvements for nginx config

matks:
| Questions     | Answers
| Description?  | Add little improvements for nginx config
| Fixed ticket? |

### DIFF
--- a/src/content/1.7/basics/installation/nginx.md
+++ b/src/content/1.7/basics/installation/nginx.md
@@ -97,7 +97,7 @@ server {
         add_header X-Content-Type-Options nosniff;
     }
 
-    # Force files inupload directory to be downloaded
+    # Force files in upload directory to be downloaded
     location ~ ^/upload/ {
         add_header Content-Disposition Attachment;
         add_header X-Content-Type-Options nosniff;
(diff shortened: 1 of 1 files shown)


src/content/1.7/basics/installation/nginx.md
```diff
@@ -157,7 +157,7 @@ server {
         deny all;
     }
     # Source code directories
-    location ~ ^/(app|bin|cache|classes|config|controllers|docs|localization|override|src|tests|tools|translations|travis-scripts|vendor|var)/ {
+    location ~ ^/(app|bin|cache|classes|config|controllers|docs|localization|override|src|tests|tests-legacy|tools|translations|travis-scripts|vendor|var)/ {
```
matks, line 160: I guess this line aims to protect installs that have been done using the source code instead of the ZIP build (because `tests` are removed in the ZIP build). So I complete it.

I thought about removing `travis-scripts` but it's better to keep it in case someone uses this configuration example for an old PrestaShop 1.7 version.

PierreRambaud (approved):

PierreRambaud:
Thank you @matks

PR #872, opened 2021-02-23
mod argument requires single quotes

davidglezz:
| Questions     | Answers
| Description?  | It took me an hour to realize that my strings were not recognized by the translation system because I used double quotes.
| Fixed ticket? | No

If we write `{l s='Welcome to this page!' mod="mymodule"}` then this string can not be translated because double quotes are not recognized.

### DIFF
--- a/src/content/1.7/modules/creation/module-translation/classic-system.md
+++ b/src/content/1.7/modules/creation/module-translation/classic-system.md
@@ -130,6 +130,10 @@ For instance, translating the string "Welcome to this page!" can be done like th
 {l s='Welcome to this page!' mod='mymodule'}
 ```
 
+{{% notice info %}}
+The `mod` argument value must go with simple quotes. Double quotes are not recognized.
+{{% /notice %}}
+
 In our sample module, the `mymodule.tpl` file...
 
 ```html


davidglezz:
... and now it is a "feature", since it was released in some stable versions.

También sería útil indicar el rango de versiones, pero para los desarrolladores de módulos esto ya es un requisito si quieren ser compatibles con todas las versiones 1.7.

I have not checked but the bug may not have been fixed

PierreRambaud:
> ... and now it is a "feature", since it was released in some stable versions.
> 
> It would also be useful to indicate the range of versions, but for module developers this is already a requirement if they want to be compatible with all 1.7 versions.
> 
> I have not checked but the bug may not have been fixed

I'm really not sure it's a feature since you can't use double quotes and force you to use a single quote, we should try to fix it instead of documenting an undesired behavior :sweat_smile:

davidglezz:
I need time to investigate the affected versions and modify the alert message to clarify that due to a bug, in several versions double quotes are not supported, and that single quotes are recommended for greater compatibility.

matks:
Hello, I agree with @PierreRambaud

`{l s='Welcome to this page!' mod="mymodule"}`
⬆️ 
this should work, unless there is a very good reason (that we do not know about)

What should be done is
1. Submit a bug report about this, verify the issue and which versions are impacted
2. See the reason why it does not work
3. Maybe, if many people have troubles with it, add a section in https://devdocs.prestashop.com/1.7/faq/ that would redirect to the bug report where all needed informations would be available

I close this PR as we will not merge it

PR #888, opened 2021-03-12
Harmonize project name

okom3pom:
| Questions     | Answers
| Description?  | Harmonisation PrestaShop


matks (approved):
Hi @okom3pom I approve this PR because there if we compare the doc before and after the PR, indeed it looks nicer _after_ the PR.

However be aware that some developers prefer to avoid (and reject) PRs that only bring "superficial" changes like this 😉  .

It's a matter of opinion of course 😅  but in the past we've got some spamming of superficial PRs which resulted in an overload of review and the merge of mistakes.

matks:
I also improved the PR title 😉

matks:
Thank you @okom3pom

PR #895, opened 2021-03-18
Fix prestashop/prestashop-webservice-lib name

industria01:
Prestashop/Prestashop-webservice-lib is invalid, it should not contain uppercase characters.


matks (approved):

matks:
Thank you @industria01

PR #903, opened 2021-03-26
Add search tip

eternoendless:
| Questions     | Answers
| Description?  | Add tip for people who struggle finding specific hooks that are described in a generic way

### DIFF
--- a/src/content/1.7/modules/concepts/hooks/list-of-hooks.md
+++ b/src/content/1.7/modules/concepts/hooks/list-of-hooks.md
@@ -20,7 +20,9 @@ A couple of hooks were modified between 1.7.0.x and 1.7.1.x.
 ## Full list
 
 {{% notice tip %}}
-**Tip:** Some hooks whose names include a page or component name are described generically using placeholders. If you can't find a hook like `actionAdminCustomersFormModifier`, try searching for `FormModifier`. 
+**Search tip:** Some hooks are generated dynamically, so their names are documented in a generic way.
+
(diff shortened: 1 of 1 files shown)


NeOMakinG (approved):

PululuK (approved):

src/content/1.7/modules/concepts/hooks/list-of-hooks.md
```diff
@@ -19,6 +19,10 @@ A couple of hooks were modified between 1.7.0.x and 1.7.1.x.
 
 ## Full list
 
+{{% notice tip %}}
+**Tip:** Some hooks whose names include a page or component name are described generically using placeholders. If you can't find a hook like `actionAdminCustomersFormModifier`, try searching for `FormModifier`. 
+{{% /notice %}}
+
```
PululuK, line 25: maybe we should clarify that these are the dynamic hooks, which is why we can list them all, which is why we have written them in a generic way.

eternoendless:
```suggestion
{{% notice tip %}}
**Search tip:** Some hooks are generated dynamically, so their names are documented in a generic way.

For example, `actionAdminCustomersFormModifier` is documented as `action<AdminControllerClassName>FormModifier`, so you won't find it if you search for the exact name. When you see a controller name or action in the hook name and you can't find it, try searching for a part of the hook name, like `FormModifier`.
{{% /notice %}}

```

PululuK: Yeah ✌🏿

eternoendless:
Thanks for the review!

PR #904, opened 2021-03-26
Update hugo to 0.82

eternoendless:
| Questions     | Answers
| Description?  | This change updates hugo version and [enables support for css classes after markdown blocks](https://gohugo.io/news/0.81.0-relnotes/#attribute-lists-after-markdown-blocks)

### DIFF
--- a/bin/installHugo.sh
+++ b/bin/installHugo.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-export HUGO_VERSION="0.76.5"
+export HUGO_VERSION="0.82.0"
 
 if [ `uname` = "Linux" ]; then
     export HUGO_OS="Linux-64bit"
--- a/src/config.yml
+++ b/src/config.yml
@@ -41,6 +41,9 @@ markup:
   goldmark:
     renderer:
       unsafe: true
+    parser:
+      attribute:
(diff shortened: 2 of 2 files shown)


PierreRambaud (approved):

Progi1984:
Please update Github Actions : https://github.com/PrestaShop/docs/blob/master/.github/workflows/build.yml#L29

PierreRambaud:
> Please update Github Actions : https://github.com/PrestaShop/docs/blob/master/.github/workflows/build.yml#L29

We should use the installHugo.sh script instead of a custom github action :thinking:

Progi1984:
We should use a github action and remove the installHugo.sh

PierreRambaud:
> We should use a github action and remove the installHugo.sh

Nope because we also need to be able to download the same hugo version without having to fetch the whole internet. We are not able to use `go dep`, and only the binary file is needed.

eternoendless:
Fixed GA on master branch

matks:
> > Please update Github Actions : https://github.com/PrestaShop/docs/blob/master/.github/workflows/build.yml#L29
> 
> We should use the installHugo.sh script instead of a custom github action 🤔

I approve ⬆️ 

Just struggled because I had hugo v0.80 😅

PR #906, opened 2021-03-26
Use correct hugo version in GitHub Actions

PierreRambaud:
| Questions     | Answers
| Description?  | Use the same hugo version as it is provided by the installHugo script

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
         working-directory: src
 
       - name: Make assets
-        run: npm install && npm run build && cd ../../../
+        run: npm install && npm run build
         working-directory: src/themes/hugo-theme-learn/_src
 
       - name: Build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: 
+on:
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -24,24 +24,24 @@ jobs:
           node-version: '12'
 
       - name: Setup Hugo
-        uses: peaceiris/actions-hugo@v2
-        with:
-          hugo-version: '0.76.5'
+        run: ../bin/installHugo.sh
```
Progi1984, line 27: Why do we pass by a script rather a github actions on this repository ?

I don't understand. We use on all our CI multiple setup actions (node, hugo, apache, mysql, node, php, python, google).

PierreRambaud: Because you can't compare node, or php with hugo.
Hugo is a software, exactly like php-cs-fixer. And we are not using php-cs-fixer github actions because we want to rely on the same version as we define in the composer.json to avoid duplicate.

Progi1984: But Hugo is comparable to Apache or MySQL.

PierreRambaud: Nope, it's comparable to Symfony :)

PierreRambaud: or if you prefer, Django or RoR :)

Progi1984: Sorry @PierreRambaud. For one time, I disapprove. I prefer use of Github Actions (more simple, more clear).

atomiix: I'm not against any of the 2 solutions but if we chose @PierreRambaud's one, we should update the docker-compose as well because I think right now it's using the `master` branch of hugo.

PierreRambaud: Three versions to maintain, not sure it's the best idea :p

PierreRambaud:
I'm closing it, the file is in conflict.

PR #930, opened 2021-04-12
Fix native version for blockreassurance

okom3pom:
| Questions     | Answers


Progi1984 (approved):

atomiix:
I'm not sure about that, it looks like the [first version released](https://github.com/PrestaShop/blockreassurance/blob/v1.0.1/blockreassurance.php#L49) was only compatible with PrestaShop 1.7+

okom3pom:
@atomiix you have right : blockreinsurance != blockreassurance

PR #908, opened 2021-03-26
Build PRs using GH actions

eternoendless:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | Pull requests weren't being built, making it possible to merge changes that would break the build

PR #958, opened 2021-04-23
WIP : Reorder the steps of launching the docs website locally

arouiadib:
When you try to start the hugo server before building the theme, a huge list of errors like :
`Failed to add template "theme/partials/header.html" in path .... : function "hugo" not defined` appears.

After installing node dependencies and building the theme, the server runs smoothly without errors.


eternoendless:
Thank you @arouiadib but this is no longer needed after #908

PR #977, opened 2021-05-04
Move Module Configuration doc to Component Storage doc page

matks:
| Questions     | Answers
| Description?  | Move Module Configuration doc to Component Storage doc page because the Configuration class is used by the Core and the Module. Add a message in Module page to redirect users to new Core page.
| Fixed ticket? | Fixes https://github.com/PrestaShop/docs/issues/954

### Screenshot

Capture d’écran 2021-05-04 à 14 57 13


PululuK (changes requested):
Improve https://devdocs.prestashop.com/1.7/modules/concepts/configuration/ instead !

matks:
> Improve https://devdocs.prestashop.com/1.7/modules/concepts/configuration/ instead !

@PululuK It's not the same?

Configuration can be used in Core too, it's not just for modules 😄  . The page https://devdocs.prestashop.com/1.7/modules/concepts/configuration/ is in Modules section.

I think it's OK to have
- 1 page that explains the component and how it is built (this page)
- 1 page that focus on "how to use it in a module"

matks:
Also https://devdocs.prestashop.com/1.7/modules/concepts/configuration/ is in "Extension Concepts" section. Configuration is not only an Extension Concept although it is very useful for modules.

Actually it is not even an Extension Concept: nobody customizes/extends the Configuration component. It is used, not extended, not completed, not decorated. It's a "Module tool". Unlike hooks.

PululuK:
@matks  in the current documentation first paragraph specifies although this concept is not only related to the module
image

So i said, instead of adding a new page and duplicating the documentation, i suggest improving the existing one :
- add details and examples of use in the core
- move to a more generic section instead of the modules section
- ...

Thanks

matks:
> @matks in the current documentation first paragraph specifies although this concept is not only related to the module
> image
> 
> So i said, instead of adding a new page and duplicating the documentation, i suggest improving the existing one :
> 
> * add details and examples of use in the core

I think the Modules page explains quite all the regular usages.

What it lacks IMO is the "what's behind this". It explains "how to use it" but not "how it's built" and both are important (although the 1st is more for modules developer and the 2nd is for maintainers and project contributors)

> * move to a more generic section instead of the modules section

I can do that and keep the old URL 👍

matks:
@PululuK After sleeping, I noticed one issue.

If I am a module developer I will likely open the "Modules" section and browse it to find what I can use.
If I am a project contributor, I will likely open the "Components" or "Architecture" section and browse that to find what I can use.

Configuration can be used by both group. But if we keep a single doc page, one of the group might miss the information.

So I suggest
1. Move current page inside Core components
2. Create small page in Modules section, to replace gone page. This small page just says "Modules can use Configuration it's very useful, see {link}"

kpodemski:
@matks 

Good idea, overall linking from the Modules page to Core stuff seems to be a best choice

matks:
WIP - I have some files to clean 😅

matks:
Ready for review ❤️

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+Secondly, modules might sometimes need a simple storage system, and they should not have to create a new SQL table for such a simple need.
+
+To address these needs, the Configuration storage of PrestaShop uses a SQL table `ps_configuration` and allows simple fetch, update and deletion of values in it.
+
+{{< figure src="/1.7/development/components/img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
+
+Each value stored in the Configuration component can be controlled through given key, is compatible with Multishop (as it can be overriden for a shop or a group of shop) and has timestamps.
+
+The `Configuration` class is the interface allowing each module developer to use the component: to get or store data.
```
PierreRambaud, line 20: We should prefer the usage of `PrestaShop\PrestaShop\Àdapter\Configuration` which is integrating the multishop behavior, no? :thinking:

matks: It depends if this documentation aims to explain "what is being used" or "what you should use" I guess

Maybe I can add a comment about this? "You will see the class Configuration used in many places but now please rather use the adapter" ?

PierreRambaud: I think it's a good idea, but this Configuration is not working well with multistore context, we also have to talk about it.

matks: > I think it's a good idea, but this Configuration is not working well with multistore context, we also have to talk about it.

What do you mean? I thought it was one of the piece of multistore that was actually working quite correctly 😅  (compared to other stuff like BO Product page)

PierreRambaud: Oh my bad, the multistore configuration is working well, but afaik, we have to prefer the usage of Adapter instead of native classes. That's the only point :sweat_smile:

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
+---
+title: Configuration storage
+weight: 80
+---
+
+# Configuration storage
+
+The Configuration storage of PrestaShop aims to address two needs.
+
+First, there is multiple situations where a single value has to be stored and be available easily. It can be a boolean to control order mode, it can be a limit for timeout: a simple key => value(s) storage is needed to manage multiple settings in PrestaShop.
```
kpodemski, line 10:
```suggestion
First, there are multiple situations where a single value has to be stored and be available easily. It can be a boolean to control order mode, it can be a limit for timeout: a simple key => value(s) storage is needed to manage multiple settings in PrestaShop.
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+Secondly, modules might sometimes need a simple storage system, and they should not have to create a new SQL table for such a simple need.
+
+To address these needs, the Configuration storage of PrestaShop uses a SQL table `ps_configuration` and allows simple fetch, update and deletion of values in it.
+
+{{< figure src="/1.7/development/components/img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
+
+Each value stored in the Configuration component can be controlled through given key, is compatible with Multishop (as it can be overriden for a shop or a group of shop) and has timestamps.
```
kpodemski, line 18:
```suggestion
Each value stored in the Configuration component can be controlled through a given key, is compatible with Multishop (as it can be overridden for a shop or a group of shops) and has timestamps.
```

src/content/1.7/development/components/configuration.md
kpodemski: I agree with @PierreRambaud that we should encourage people to use an Adapter, especially that it's working fine with multi-store now

kpodemski: maybe we need to have this as a separate paragraph? to inject adapter through constructor or create an instance of it, because there are no static methods in it

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Store configuration data
+
+```php
+<?php
+Configuration::updateValue(string $key, mixed $value);
+```
+
+`$key` will be the reference to your data to reuse later.
+`$value` can be any scalar type (int, string, bool, array...).
```
kpodemski, line 32: Most of the time array value means that configuration is multi-lang, we need to point this because if someone has a structure like:

```
0 => 'XYZ',
1 => 'ABC,
```

it will be treated as multi-lang even though the developer might not need it

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+With `$keys` as an array of keys to retrieve.
+
+This returns an array, containing the values stored in the configuration table or null if a key does not exist.
+
+Example:
+```php
+<?php
+Configuration::getMultiple(array('PS_VERSION_DB', 'UNKNOWN_KEY'));
```
kpodemski, line 78:
```suggestion
Configuration::getMultiple(['PS_VERSION_DB', 'UNKNOWN_KEY']);
```

I think our CS is setting arrays to short version right?

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+// returns this array:
+array(
+  'PS_VERSION_DB' => '1.7.4.0',
+  'UNKNOWN_KEY' => null,
+)
```
kpodemski, line 84:
```suggestion
[
  'PS_VERSION_DB' => '1.7.4.0',
  'UNKNOWN_KEY' => null,
]
```

kpodemski (changes requested):
Small feedback

src/content/1.7/development/components/configuration.md
eternoendless: We are supposed to encourage the use of interfaces, not adapters: remember, adapters are meant to go away ;)
...but that would require one interface for each class

src/content/1.7/development/components/configuration.md
eternoendless: Also, an array is not a scalar:

> Scalar variables are those containing an int, float, string or bool. Types array, object and resource are not scalar.

https://www.php.net/manual/en/function.is-scalar.php

eternoendless: Follow up from @kpodemski 's comment
```suggestion
`$value` can be any scalar type (int, float, string, bool) or an array.

{{% notice tip %}}
**Multi language**.

By convention, multilanguage values are stored in arrays indexed by language id, like this:

```php
[
    123 => 'Value in some language',
    456 => 'Value in some other language',
]
```
{{% /notice %}}
```

File: src/content/1.7/development/components/configuration.md (hunk 1 above)
Comment (kpodemski) [reply src/content/1.7/development/components/configuration.md:20]: @eternoendless I was not precise enough, in almost every new code in PrestaShop, where we need, we're injecting @prestashop.adapter.legacy.configuration' 

it has a proper interface and... this is recommended way, right? 🤔

File: src/content/1.7/development/components/configuration.md [hunk 7]
```diff
@@ -0,0 +1,93 @@
...
+Each value stored in the Configuration component can be controlled through given key, is compatible with Multishop (as it can be overriden for a shop or a group of shop) and has timestamps.
+
+The `Configuration` class is the interface allowing each module developer to use the component: to get or store data.
+
+## Usage
+
+### Store configuration data
+
+```php
+<?php
```
eternoendless, line 27: No longer necessary
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+Configuration::updateValue(string $key, mixed $value);
+```
+
+`$key` will be the reference to your data to reuse later.
+`$value` can be any scalar type (int, string, bool, array...).
+
+### Check a configuration data set exists
+
+```php
+<?php
```
eternoendless, line 37:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+
+`$key` is the name of the configuration data to check. The method returns `true`
+if the data exists, `false` otherwise.
+
+### Retrieve configuration data
+
+#### Single key
+
+```php
+<?php
```
eternoendless, line 49:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+Configuration::get(string $key);
+```
+
+With `$key` as the data to retrieve.
+If the key does not exist, the returned value will be null.
+
+Example:
+```php
+<?php
```
eternoendless, line 58:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+Configuration::get('PS_VERSION_DB');
+
+// returns '1.7.4.0'
+```
+
+#### Multiple keys
+
+```php
+<?php
```
eternoendless, line 67:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+Configuration::getMultiple(array $keys);
+```
+
+With `$keys` as an array of keys to retrieve.
+
+This returns an array, containing the values stored in the configuration table or null if a key does not exist.
+
+Example:
+```php
+<?php
```
eternoendless, line 77:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,93 @@
...
+// returns this array:
+array(
+  'PS_VERSION_DB' => '1.7.4.0',
+  'UNKNOWN_KEY' => null,
+)
+```
+#### Delete a key
+
+```php
+<?php
```
eternoendless, line 89:
```suggestion
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,86 @@
...
+
+The Configuration storage of PrestaShop aims to address two needs.
+
+First, there are multiple situations where a single value has to be stored and be available easily. It can be a boolean to control order mode, it can be a limit for timeout: a simple key => value(s) storage is needed to manage multiple settings in PrestaShop.
+
+Secondly, modules might sometimes need a simple storage system, and they should not have to create a new SQL table for such a simple need.
+
+To address these needs, the Configuration storage of PrestaShop uses a SQL table `ps_configuration` and allows simple fetch, update and deletion of values in it.
+
+{{< figure src="/1.7/development/components/img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
```
eternoendless, line 16: You can use relative paths
```suggestion
{{< figure src="img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
```

src/content/1.7/development/components/configuration.md
eternoendless: Yep, that's the recommended way, depend on the interface, inject the adapter

src/content/1.7/modules/concepts/configuration.md
```diff
@@ -5,81 +5,6 @@ weight: 1
...
-#### Delete a key
-
-```php
-<?php
-Configuration::deleteByName(string $key);
-```
-`$key` is the name of the configuration data to delete. The method returns `true`
-if the key is removed, `false` otherwise.
+{{% notice note %}}
+The content of this page has been moved to [Core Reference > Components > Configuration storage](/1.7/development/components/configuration/).
```
eternoendless, line 9: Use refs or there will be dead links without us noticing
```suggestion
The content of this page has been moved to [Core Reference > Components > Configuration storage]({{< ref "/1.7/development/components/configuration" >}}).
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Configuration storage
+weight: 80
```
eternoendless, line 3:
```suggestion
weight: 80
aliases:
  - /1.7/modules/concepts/configuration/
```

matks:
@eternoendless I see that you have added commits on my branch and deleted the old page _but_ you also added an alias so people that follow the old URL will find the new page.

However as I wrote

> If I am a module developer I will likely open the "Modules" section and browse it to find what I can use.
If I am a project contributor, I will likely open the "Components" or "Architecture" section and browse that to find what I can use.
>
> Configuration can be used by both group. But if we keep a single doc page, one of the group might miss the information.
>
> So I suggest
>
> Move current page inside Core components
Create small page in Modules section, to replace gone page. This small page just says "Modules can use Configuration it's very useful, see {link}"

I kept both pages on purpose, so that people that are navigating inside the "Module" section can find it. Do you think it was a bad idea?

eternoendless:
I think core components should ALL be explained in the Core section

matks:
@eternoendless @kpodemski My apologies, I fixed an issue and I rebased my PR which has deleted all your commits. I re-copied your suggestions, finished the last things you suggested and finally put it all in a single commit.

matks:
> I think core components should ALL be explained in the Core section

https://github.com/PrestaShop/docs/issues/1058

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,103 @@
...
+
+The Configuration storage of PrestaShop aims to address two needs.
+
+First, there are multiple situations where a single value has to be stored and be available easily. It can be a boolean to control order mode, it can be a limit for timeout: a simple key => value(s) storage is needed to manage multiple settings in PrestaShop.
+
+Secondly, modules might sometimes need a simple storage system, and they should not have to create a new SQL table for such a simple need.
+
+To address these needs, the Configuration storage of PrestaShop uses a SQL table `ps_configuration` and allows simple fetch, update and deletion of values in it.
+
+{{< figure src="../img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
```
eternoendless, line 18: Don't you think it would be better to store this within the components section? (If yes, you will need to move the file)

```suggestion
{{< figure src="img/configuration_sql_structure.png" title="Configuration SQL structure" >}}
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,103 @@
...
+
+Each value stored in the Configuration component can be controlled through a given key, is compatible with Multishop (as it can be overridden for a shop or a group of shops) and has timestamps.
+
+The storage is performed in `Configuration` class.
+
+We recommend to access it through service `prestashop.adapter.legacy.configuration` (class `PrestaShop\PrestaShop\Adapter`), this is the interface allowing each developer to get or store configuration data.
```
eternoendless, line 24:
```suggestion
The storage is managed by the `Configuration` class.

{{% notice tip %}}
Although you can access `Configuration` statically from any class, we recommend injecting the `prestashop.adapter.legacy.configuration` service instead.
{{% /notice %}}
```

src/content/1.7/development/components/configuration.md
```diff
@@ -0,0 +1,103 @@
...
+```php
+Configuration::updateValue(string $key, mixed $value);
+```
+
+`$key` will be the reference to your data to reuse later.
+`$value` can be any scalar type (int, float, string, bool) or an array.
+
+{{% notice tip %}}
+
+**Multi language**.
```
eternoendless, line 39: Remove extra line break
```suggestion
**Multi language**.
```

src/content/1.7/development/components/configuration.md
matks: I'm not to understand, the PNG image is already inside `components` section.
```
src/content/1.7/development/components/img/configuration_sql_structure.png 
```

To be able to use `img/configuration_sql_structure.png` filepath we would need to create a `configuration` subdirectory inside `component` directory, and its only purpose would be to contain `img`.

is it worth it?

eternoendless: Oh right, forgot it wasn't a directory.

eternoendless (approved):

eternoendless:
Thank you @matks

Issue #905, opened 2021-03-26
CQRS validation


velorb:
I have question about your approach to validation.
I've discovered validation rules in: forms, commands before dispatch, command handlers, and entities (eg. when add product to cart). This includes a lot of repetition.

Can anyone describe how validation should be performed?
- Where validate data rules (like max length, correct syntax)?
- Where validate business rules?
- How catching business exceptions and forwarding messages from them to symfony form supports assync command handling?

I read about future plans to move all backend/frontend presentation to vue. How symfony forms will be connected with vue?  Does creating symfony forms make sens with vue in mind? Shouldn't form validation will be performed in commands?

matks:
> I've discovered validation rules in: forms, commands before dispatch, command handlers and entities (eg. when add product to cart). This includes a lot of repetition.

Hello, this is wanted. These are all different levels of validation, and you might want to handle them differently.

**Form level**

We want to validate the data submitted in the form is correct by itself. For example if a form input expects `Yes` or `No` and user submitted `A` we want to reject the data 😄  . This first level of validity aims to validate "user filled and submitted the form correctly"

If you catch an issue at form level you can tell user using an error message attached to the input used 👍  .

_So we dont look at "is the data valid" but rather "is the form submitted correctly, according to its settings and validation rules?"_

**Before dispatch ; Data validation**

At this level we aim to validate, when possible, the data itself. Not the form, we forget about the form. For example if user has submitted a Country ISO code that does not exist, we can detect it here. We also check data type. However we try not to perform too many SQL queries so sometimes we validate a data later if it's very expensive to do it in this level.

We do this validation because, as much as possible, we want to NOT dispatch a command with not valid data. If we can detect data is not valid before dispatch, we try to do it here.

If you catch an issue at data level you can tell user using an error message on the BO page 👍  . You dont know what form input submitted it but at least you can tell merchant "hello ! we did NOT performed your action because some data is not valid"

_So we look at "is the data valid" but not the business rules._

**Command handlers**

At this level we aim to validate business rules, we WANT to perform the action required by the user. But when performing the action, we might find issues, and you can only find them when performing heavy SQL statements. For example I have a Cart ID: 8278. I perform `getCartById($myId);` but it fails and returns a boolean. So I cannot perform the action.

If you catch an issue at command handler level, you have to throw an Exception. This is less user friendly althought you can display them nicely by catching them at controller level. Also maybe some data has been modified then process was halted so it's bad we found the data is not valid so late.

[...]

PR #1000, opened 2021-05-18
Explain validation flow in migration guide

zuk3975:
| Questions     | Answers
| Description?  | explain validation flow. Form -> Commands (->Value objects) -> CommandHandlers (->Domain services).
| Fixed ticket? | #905


src/content/1.7/development/architecture/domain/cqrs.md
```diff
@@ -30,6 +30,20 @@ Fortunately, by implementing CQRS it allows PrestaShop to quickly build new API,
 
 {{< figure src="../img/new_architecture.png" title="New architecture using CQRS" >}}
 
+## Validation
```
zuk3975, line 33: Im not sure its the best way... 
Maybe this needs a separate page, which would describe the flow of all the validation instead? (just idk where to put it).

src/content/1.7/development/architecture/migration-guide/validation.md
```diff
@@ -0,0 +1,18 @@
...
+* **Entities** (or the `ObjectModels` in Prestashop) - This is the last level. It is there to prevent data integrity issues (like attempting to store a string into an integer column). This is something that, if all data validations before were done correctly, should rarely happen. It's like **the very last protection against nasty data being persisted**. If these errors still occur, you might get a legacy `PrestaShopException`, but in recent migration process we try to catch it early and replace it by related `CoreException` or `DomainException`. See more in [Domain services]({{< relref "/1.7/development/architecture/domain/domain-services" >}}).
```
matks, line 12:
```suggestion
* **Entities** (or the `ObjectModels` in Prestashop) - This is the last level. It is there to prevent data integrity issues (like attempting to store a string into an integer column). This is something that, if all data validations before were done correctly, should rarely happen. This is the last step before data is being persisted: if invalid data has slipped through validation levels and passes this step, then it is persisted in database and might introduce inconsistancy in the application. If these errors still occur, you might get a legacy `PrestaShopException`, but in recent migration process we try to catch it early and replace it by related `CoreException` or `DomainException`. See more in [Domain services]({{< relref "/1.7/development/architecture/domain/domain-services" >}}).
```
I don't know if 'nasty' is good looking in a documentation 😅

matks (approved):
Looks good, added 1 suggestion

src/content/1.7/development/architecture/migration-guide/validation.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Validation
+weight: 70
+---
+
+# Validation
+
+The purpose of validation is to protect the application from failures, state inconsistency, security related issues etc. Prestashop has different levels of validation:
```
kpodemski, line 8:
```suggestion
The purpose of validation is to protect the application from failures, state inconsistency, security related issues etc. PrestaShop has few different levels of validation:
```

src/content/1.7/development/architecture/migration-guide/validation.md
```diff
@@ -0,0 +1,18 @@
...
+* **Forms** (the [CRUD]({{<relref "/1.7/development/architecture/migration-guide/forms/crud-forms">}}) or [Settings]({{<relref "/1.7/development/architecture/migration-guide/forms/settings-forms">}})) - are using [Symfony validation constraints](https://symfony.com/doc/3.4/validation.html#constraints) to prevent the user from providing invalid information while also enriching the user experience. For example if a form input expects `Yes` or `No`, but user submitted value `XYZ` we want to reject the data and inform the user what was wrong. The best approach (if possible) is to prevent user providing the wrong value in first place, so in this example we could simply use a checkbox or select input with only 2 possible selections. There are some custom constraints implemented in Prestashop [here](https://github.com/PrestaShop/PrestaShop/tree/1.7.8.x/src/Core/ConstraintValidator).
```
kpodemski, line 9:
```suggestion
* **Forms** (the [CRUD]({{<relref "/1.7/development/architecture/migration-guide/forms/crud-forms">}}) or [Settings]({{<relref "/1.7/development/architecture/migration-guide/forms/settings-forms">}})) - are using [Symfony validation constraints](https://symfony.com/doc/3.4/validation.html#constraints) to prevent the user from providing invalid information while also enriching the user experience. For example if a form input expects `Yes` or `No`, but user submitted value `XYZ` we want to reject the data and inform the user what was wrong. The best approach (if possible) is to prevent user providing the wrong value in the first place, so in this example we could simply use a checkbox or select input with only 2 possible selections. There are some custom constraints implemented in PrestaShop [here](https://github.com/PrestaShop/PrestaShop/tree/1.7.8.x/src/Core/ConstraintValidator).
```

src/content/1.7/development/architecture/migration-guide/validation.md
```diff
@@ -0,0 +1,18 @@
...
...ecture/domain/value_objects" >}}) - after the form is submitted and validated, controller is usually dispatching some `Command`. We must ensure that commands are valid too, because in theory, commands should not be aware whether it was dispatched from a controller or from a command line, therefore it is possible that inputs are not yet validated by the form. The `command` itself should not handle the validation (although there are some older commands that does that, but that might only be some early migration mistakes), but **use the value objects in the constructor instead**. Value objects has their own validation rules and can be reused in any other command. For example, take a look at [ProductCondition](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/src/Core/Domain/Product/ValueObject/ProductCondition.php) - this is a value object of `Product` domain. It allows constructing specific values and throws `ProductConstraintException` if invalid one is provided. It is currently used in [UpdateProductOptionsCommand](https://github.com/PrestaShop/PrestaShop/1.7.8.x/develop/src/Core/Domain/Product/Command/UpdateProductOptionsCommand.php) and helps to maintain the command validity.
```
kpodemski, line 10:
```suggestion
* [Value objects]({{< relref "/1.7/development/architecture/domain/value_objects" >}}) - after the form is submitted and validated, controller is usually dispatching some `Command`. We must ensure that commands are valid too, because in theory, commands should not be aware whether they were dispatched from a controller or from a command line, therefore it is possible that inputs are not yet validated by the form. The `command` itself should not handle the validation (although there are some older commands that does that, but that might only be some early migration mistakes), but **use the value objects in the constructor instead**. Value objects have their own validation rules and can be reused in any other command. For example, take a look at [ProductCondition](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/src/Core/Domain/Product/ValueObject/ProductCondition.php) - this is a value object of `Product` domain. It allows constructing specific values and throws `ProductConstraintException` if invalid one is provided. It is currently used in [UpdateProductOptionsCommand](https://github.com/PrestaShop/PrestaShop/1.7.8.x/develop/src/Core/Domain/Product/Command/UpdateProductOptionsCommand.php) and helps to maintain the command validity.
```

matks (approved):

matks:
Requested changes have been applied so thank you @zuk3975

PR #1016, opened 2021-05-26
Update migration.md

GabrieleMartini:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | Fix typo
| Fixed ticket? | Fixes #{issue number here} if there is a related issue

PR #1017, opened 2021-05-26
Update migration.md

GabrieleMartini:
| Questions     | Answers
| Description?  | Fix typo

### DIFF
--- a/src/content/1.7/basics/keeping-up-to-date/migration.md
+++ b/src/content/1.7/basics/keeping-up-to-date/migration.md
@@ -406,7 +406,7 @@ to change as shown on the screenshot.
 * Start by selecting your CSV file.
 * The second option to modify will be the format. Choose “CSV” or “CSV for MS Excel”.
 * Skip the first line of your file. We had the column names in the first line to make our file edition easier, but we do not want to import it in the database.
-* Disable the foreign key checks. We know some relations don’t exist yet, because they won’t be imported at the same time.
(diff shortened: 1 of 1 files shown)


PululuK (changes requested):
See https://github.com/PrestaShop/docs/pull/1016#pullrequestreview-668840752

src/content/1.7/basics/keeping-up-to-date/migration.md
```diff
@@ -406,7 +406,7 @@ to change as shown on the screenshot.
 * Start by selecting your CSV file.
 * The second option to modify will be the format. Choose “CSV” or “CSV for MS Excel”.
 * Skip the first line of your file. We had the column names in the first line to make our file edition easier, but we do not want to import it in the database.
-* Disable the foreign key checks. We know some relations don’t exist yet, because they won’t be imported at the same time.
+* Disable the foreign key checks. We know some relations doesn’t exist yet, because they won’t be imported at the same time.
```
matks, line 409:
```suggestion
* Disable the foreign key checks. We know some relations do not exist yet, because they won’t be imported at the same time.
```
Some relations is plural so `do not` instead of `does not`

matks (changes requested):

src/content/1.7/basics/keeping-up-to-date/migration.md
GabrieleMartini: You're right

PR #1021, opened 2021-05-26
Update migration.md

GabrieleMartini:
| Questions     | Answers

### DIFF
--- a/src/content/1.7/basics/keeping-up-to-date/migration.md
+++ b/src/content/1.7/basics/keeping-up-to-date/migration.md
@@ -593,7 +593,7 @@ missing product to the index”.
 
 The manual process explained to export and import data with CSV files
 from database. If you are a developer and you are willing to build your
-own tools for migrations, other ways exist to get the business data.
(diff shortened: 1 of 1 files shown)


PululuK (changes requested):
See https://github.com/PrestaShop/docs/pull/1016#pullrequestreview-668840752a

src/content/1.7/basics/keeping-up-to-date/migration.md
```diff
@@ -593,7 +593,7 @@ missing product to the index”.
 
 The manual process explained to export and import data with CSV files
 from database. If you are a developer and you are willing to build your
-own tools for migrations, other ways exist to get the business data.
+own tools for migrations, other ways exists to get the business data.
```
matks, line 596: There are many "other ways" so plural is expected

matks (changes requested):

matks:
@GabrieleMartini Thank you for the proof-reading but it seems this PR actually does not fix an issue 😉 so I close it

PR #1022, opened 2021-05-26
Update upgrade.md

GabrieleMartini:
| Questions     | Answers
| Description?  | Fix typo

### DIFF
--- a/src/content/1.7/basics/keeping-up-to-date/upgrade.md
+++ b/src/content/1.7/basics/keeping-up-to-date/upgrade.md
@@ -213,7 +213,7 @@ Modules page, under the tab “Notifications”:
 
 ## 1-click upgrade module
 
-If the manual process seems too tedious, other solutions exist to complete an upgrade.
+If the manual process seems too tedious, other solutions exists to complete an upgrade.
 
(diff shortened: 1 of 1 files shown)


src/content/1.7/basics/keeping-up-to-date/upgrade.md
```diff
@@ -213,7 +213,7 @@ Modules page, under the tab “Notifications”:
 
 ## 1-click upgrade module
 
-If the manual process seems too tedious, other solutions exist to complete an upgrade.
+If the manual process seems too tedious, other solutions exists to complete an upgrade.
```
atomiix, line 216: The plural doesn't take an "s" AFAIK 🤔

matks: Agreed with @atomiix 👍  although I understand that the context is confusing. The only "other solution" mentioned is the 1-click upgrade module, even though the doc says there is other solutionS it only highlights one 😅

matks:
@GabrieleMartini Thank you for the proof-reading but it seems this PR actually does not fix an issue 😉 so I close it

PR #1053, opened 2021-06-25
Change value of the table_open_cache to recommended 4000

### LINKED ISSUE #428: Optimizations docs page recommends a lower table_open_cache value than the default
optimizations page says to set `table_open_cache = 1000`
https://devdocs.prestashop.com/1.7/scale/optimizations/#5-mysql-mariadb-settings

while the default of MySQL is **2000** for MySQL <8.0.3 and **4000** for MySQL >8.0.4
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_table_open_cache

is this intentional or a mistake, perhaps a recommended value should be higher than the minimum default?

kpodemski:
| Questions     | Answers
| Description?  | Change is coming from a suggestion from Jocelyn:
https://github.com/PrestaShop/docs/issues/428#issuecomment-829000403
| Fixed ticket? | Fixes #428 
| Sponsor company | impSolutions

### DIFF
--- a/src/content/1.7/scale/optimizations.md
+++ b/src/content/1.7/scale/optimizations.md
@@ -176,7 +176,7 @@ As always, these values should be adjusted to your own environment, you probably
 query_cache_limit = 128K 
 query_cache_size = 32M
 query_cache_type = ON
-table_open_cache = 1000
+table_open_cache = 4000
 thread_cache_size = 80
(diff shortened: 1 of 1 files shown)


matks (approved):

matks:
Thank you @kpodemski

Issue #241, opened 2019-04-10
Improve menu navigation


sarjon:
Current menu structure is more or less weird, it seems that topics are randomly put in any location that seems "most appropriate".

Some rules should be defined:
1. What is "Core development"? What can be considered as "Core development"? What should be at the root of this section? Same for "Modules" and "Themes" sections.

My suggestion for menu (it's more like "big picture" without much details):
1. Basics
    * Introduction
    * Installation 
2. Core development
    * Standards & conventions
        * Coding standards
        * Naming conventions
    * Architecture
    * Components
        * Grid
    * Testing
        * Unit tests
        * BDD tests
        * E2E tests
    * CLDR
    * Database
    * Webservice
2. Module development
    * Getting started
        * Legacy Admin controllers
        * Symfony admin controllers
        * Front controllers
    * Testing
    * Hooks
    * Payment modules
    * Shipping modules
3. Theme development
    * Getting started
4. Contribute
    * Code
    * Documentation
5. Cookbook (Scale previously)
    * Optimization
        * How to benchmark PrestaShop?
        * How to optimize PrestaShop?

Some things that could be improved:
1. Core changes should not be under "Modules" section. Maybe there should be "Changelog" as separate section to document BC breaks & most notable changes (just an idea).
https://prnt.sc/n9xd6y

2. I think that "Documentation" section should not be part of Documentation itself. It's mainly about contributing to Docs, so I think it should be moved under "Contribute".
https://prnt.sc/n9xl50

3. There should not be "How tos" at the root of "Core development". These are very small topics thus should be under some subsection, for example "Cookbook", like:
* Cookbook
    * Email
        * How to send e-mails?

    https://prnt.sc/n9xo72

4. Side menu levels are not easy to follow. Maybe styling should be improved or maybe dropdowns added, not sure. :/
https://prnt.sc/na0xrm

sarjon:
@matks can you take a look at this issue? I was navigating through menu it for me, it's such a mess. :/ Do you think we can improve it?

matks:
@eternoendless wdyt ? 😄

tomas862:
seems better structure for me as well :+1:

sarjon:
ping :ping_pong: @eternoendless :ping_pong: pong

eternoendless:
I agree that the structure is not optimal, give me some time to think about it.

[...]

PR #1059, opened 2021-06-29
Improve introduction, move cache section to core/architecture

eternoendless:
| Questions     | Answers
| Description?  | Improved the module's introduction article, moved cache part to core/architecture because it had little to do with modules, really.

### DIFF
--- a/src/content/1.7/development/architecture/cache.md
+++ b/src/content/1.7/development/architecture/cache.md
@@ -1,0 +1,16 @@
+---
+title: Cache
+---
+
+# Cache
+
+## Classes in the root namespace and overrides
+
+The `/var/cache/<ENV>/class_index.php` file contains the link between the class and the declaration file. If there is a caching issue, this file can safely be deleted.
+
+## Smarty
+
+When the store's front-end doesn't quite reflect your changes and emptying the browser's cache is not effective, try emptying the following folders:
+
+- `/var/cache/<ENV>/smarty/cache`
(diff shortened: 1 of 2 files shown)


matks (approved):
Do you think it would be useful to add links from Modules introduction page to this new cache page, in order to help module developers that explore the Module section find this page?

I don't know whether module developers read both "Modules" section and "Core reference", I'm afraid some might skip "Core reference" (that might need to be renamed btw) and go straight to "Modules" section, thus missing out some important documentation.

eternoendless:
I think your suggestion is definitely valid, but not in the introduction. This link should probably go in the Extension concepts part. I'm going through the whole documentation right now, I'll be pushing it page by page as separate PRs as I go so it's easier to review and merge quickly.

matks:
> I think your suggestion is definitely valid, but not in the introduction. This link should probably go in the Extension concepts part. I'm going through the whole documentation right now, I'll be pushing it page by page as separate PRs as I go so it's easier to review and merge quickly.

If you go through the whole documentation, I would suggest you try to analyze how you feel about each section. Which ones are big and could be split, which ones are too small. Did you see duplicate content? Did you see 2 topics that should be merged into one ? Did you read, many pages later, some information you wish you were provided 10 minutes before ?

Basically going through the whole documentation should give you the necessary vision to [find a good new structure for the documentation](https://github.com/PrestaShop/docs/issues/241).

eternoendless:
For sure :)

matks (approved):

PR #1063, opened 2021-07-02
Update Module > Getting Started section

eternoendless:
I updated and broken down the Getting Started part by putting the Tutorial part on a separate page.

Screenshot 2021-07-02 at 16 55 22

Also:

- The Configuration Storage page has been updated and broken in two, with the service being put forward vs the Configuration class.
- Added a Module class reference page with a lot of properties described


matks (approved):

matks:
Thank you @eternoendless

PR #1067, opened 2021-07-03
Add a notice for how to use API key in header

Arman-Hosseini:
| Questions     | Answers
| Description?  |To use the API key in header, we need to encode the API key with `base64_encode`.
| Fixed ticket? | N/A


eternoendless (approved):
I simplified the notice and moved it below the example for clarity reasons.

eternoendless:
Thank you @Arman-Hosseini

PR #1073, opened 2021-07-06
Fix diverse style issues

eternoendless:
| Questions     | Answers
| Description?  | This PR basically fixes small visual issues with horizontal scrolling in code blocks.<br>While I was doing that, I realized that the theme was using the `highlight` css class to highlight search results (no longer needed because we now use Algolia, but still), and this was conflicting with the same class being used by hugo in code blocks. I replaced `highlight` with `highlight-words`.<br>Also, the copy to clipboard button is no longer attached to inline code snippets.

Preview:
Screenshot 2021-07-06 at 16 46 46

### DIFF
--- a/src/themes/hugo-theme-learn/assets/sass/_prestashop.scss
+++ b/src/themes/hugo-theme-learn/assets/sass/_prestashop.scss
@@ -34,7 +34,7 @@ a:hover {
   color: $menuVisitedColor;
 }
 
-#body a.highlight:after {
+#body a:after {
   display: block;
   content: '';
   height: 1px;
--- a/src/themes/hugo-theme-learn/assets/sass/_prestashop.scss
+++ b/src/themes/hugo-theme-learn/assets/sass/_prestashop.scss
@@ -143,6 +143,11 @@ a[data-featherlight] {
     right: 3rem;
   }
 
+  &:after {
+    // remove underline effect
+    content: none !important;
+  }
+
   > img {
     max-height: 350px;
   }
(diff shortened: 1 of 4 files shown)


src/themes/hugo-theme-learn/assets/sass/_prestashop.scss
```diff
@@ -143,6 +143,11 @@ a[data-featherlight] {
     right: 3rem;
   }
 
+  &:after {
```
NeOMakinG, line 146:
```suggestion
  &::after {
```

It's a good practice in order to make it more understandable introduced in CSS3, so we distinguish Pseudo-class and pseudo-elements

src/themes/hugo-theme-learn/assets/sass/_theme.scss
```diff
@@ -543,6 +551,13 @@ pre {
   }
 }
 
+div.highlight {
```
NeOMakinG, line 554: No chance that `highlight` class could be on another element type than a `div`?

NeOMakinG:
2 little things, overwise it looks fine to me!

src/themes/hugo-theme-learn/assets/sass/_theme.scss
eternoendless: Not impossible, but this targets precisely the div that is added by Hugo around code blocks

PierreRambaud (approved):

matthieu-rolland (approved):

eternoendless:
Thanks for the review!

PR #1078, opened 2021-07-11
Update getting-started.md

armanism24:
Added API Blog

| Questions     | Answers


kpodemski:
Sorry but I think it's not necessary.

matks:
Indeed, it seems to be promotion and not information

PR #1080, opened 2021-07-12
Add 1.7.9 version

jf-viguier:
| Questions     | Answers
| Description?  | Add 1.7.9 version php support to the doc
| Fixed ticket? | no

### DIFF
--- a/src/content/1.7/basics/installation/system-requirements.md
+++ b/src/content/1.7/basics/installation/system-requirements.md
@@ -185,6 +185,24 @@ You'll get a web page detailing requirements and recommendations, and how your s
       </td>
       <td class="support-no"><span class="sr-only">No</span></td>
     </tr>
+    <tr>
+      <td>1.7.9</td>
+      <td class="support-no"><span class="sr-only">No</span></td>
+      <td class="support-no"><span class="sr-only">No</span></td>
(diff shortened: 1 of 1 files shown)


jf-viguier:
https://github.com/PrestaShop/PrestaShop/issues/25324

matks:
Hi @jf-viguier there is no 1.7.9 scheduled. Next version will be a 8.0.0 we think, so a new major version.

We need to upgrade Symfony to version 4 and we need to be compatible with PHP8 and such changes require us to upgrade the version number following SemVer.

Don't worry (I anticipate questions), PrestaShop 8.0.0 will NOT carry a big change 😄 , unlike PrestaShop 1.7.0 that introduced Symfony. PrestaShop 8 will follow the journey of the 1.7 versions, and the continuous improvements and projects started and carried out in 1.7 versions will continue through version 8. "Major" does not mean "big", it's simply SemVer convention that pushes us to mark it as a v8.0.0 and not a 1.7.9.0

Also, the opportunity of this new major version allows us to drop the 1.x version number and comply with SemVer format: Major.Minor.Patch 🎉  something we wanted to do for a long time!

matks:
I close as this change will not be merged 😉  thank you for your contribution

PR #1087, opened 2021-07-16
Fix broken hooks list

eternoendless:
| Questions     | Answers
| Description?  | It looks like it's not possible to add notice blocks inside definition lists, it will break the definition list. Also, whitespace is important.


matks (approved):

matks:
Thank you @eternoendless

PR #1089, opened 2021-07-21
Update nginx.md

armanism24:
| Questions     | Answers
| Description?  | Added guide for installing SSL certificate on NGINX.


Progi1984:
Indeed, it seems to be promotion and not information

PR #1098, opened 2021-08-02
Added troubleshoot documentation

maldarol:
| Questions     | Answers
| Description?  | Described a problem’s solution which can occur when there is no json file for the npm build.
| Fixed ticket? | Fixes https://github.com/PrestaShop/example-modules/issues/72

### DIFF
--- a/src/content/1.7/development/compile-assets.md
+++ b/src/content/1.7/development/compile-assets.md
@@ -90,3 +90,11 @@ How we manage to work with this, as well as working on previous PrestaShop versi
 
 There are a lot of tools able to do this. At PrestaShop, we mainly use the ['n' package](https://www.npmjs.com/package/n?activeTab=versions) or [nvm](https://github.com/nvm-sh/nvm).
 This is pretty easy to use, when you are working on a directory which need an older or newer node version, use one of these tools to switch and then `npm install && npm run build`.
+
+If you working with PrestaShop's production release for an actual shop instead of development, new-theme directory will not have a package.json, this means that the dependencies will be incorrect and the `npm run build` will display an error message like this:
+
+```
+Module build failed: Error: Couldn't find preset "env" relative to directory
+```
+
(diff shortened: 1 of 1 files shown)


matks:
Hello,

I have modified your PR to target the `1.7.x` branch.

We have modified the devdocs inner working in order to welcome multiple documentation versions. This is because soon we will have a 1.7 documentation and a 8.0 documentation, matching PrestaShop 1.7 or PrestaShop 8.0 softwares.

Read more on https://devdocs.prestashop.com/1.7/contribute/documentation/how/documentation-versions/

Please also note that now that we have multiple versions inside one documentation, we had to build a workflow that can handle this setup. This repository https://github.com/PrestaShop/docs now only contains the content of the documentation, and https://github.com/PrestaShop/devdocs-site contains the sources and the deployment workflow.

maldarol:
Thank you for the help! Is there anything I should do other than resolving the conflict?

matks:
Hello again ! Sorry for late answer.

I fail to understand why someone would work with PrestaShop's production release 😅  ?

- if you want to contribute to https://github.com/prestashop/prestashop you will use the sources
- if you are building a module or a theme, you do not need to compile assets?

Why do you need to compile assets in your case 🤔  ? Assets are built for the ZIP shipped to production, and are not meant to be modified afterwards ?

maldarol:
I created this PR because I needed to compile assets for my module because the grid I created for it did not worked (the Javascript needed for the row actions was missing). With the production release the compiling always failed with an error.

Maybe there is a better way to include js files to a module but I found this solution. I was thinking that someone else might run into this so it would be good if there was documentation about it.

matks (changes requested):
Sorry for late answer. I think this should rather go in Frequently Asked Questions https://devdocs.prestashop.com/1.7/faq/ not in this part because this is not a standard use of PrestaShop.

matks:
I think this PR is stale so I close it. As the contribution was interesting I ported it to PR https://github.com/PrestaShop/docs/pull/1247

PR #1131, opened 2021-08-25
Explain where data need to be stored in module's context

PierreRambaud:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | In addition to https://github.com/PrestaShop/PrestaShop/pull/25702

### DIFF
--- a/modules/creation/good-practices.md
+++ b/modules/creation/good-practices.md
@@ -69,7 +69,7 @@ menuTitle: Good practices
 - A merchant is likely to have a shop running on a different shop than yours. In case a module is using PHP extensions not installed by default by PHP,
 add a preliminary check before using them (I.e with `extension_loaded`). This prevents fatal errors to be thrown on shops on which these extensions aren’t enabled.
 
-- About files, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scalable applications, and also to keep data if needed in case of you uninstall the module.
+- If you need to store static files: we recommend putting temporary or cache files in prestashop's `/var/cache/<env>/modules/YOUR_MODULE/` directory, and static or shared files in prestashop's  `/var/modules/YOUR_MODULE/` directory. Writing in the `/modules` directory is not recommended, because doing so will make it harder to use your module in distributed environments. In addition, by storing files outside your module's directory, they are kept even if the module is uninstalled.
 
 ## A few recommendations for your themes
 
--- a/modules/creation/good-practices.md
+++ b/modules/creation/good-practices.md
@@ -69,6 +69,8 @@ menuTitle: Good practices
 - A merchant is likely to have a shop running on a different shop than yours. In case a module is using PHP extensions not installed by default by PHP,
 add a preliminary check before using them (I.e with `extension_loaded`). This prevents fatal errors to be thrown on shops on which these extensions aren’t enabled.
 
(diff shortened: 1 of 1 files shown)


modules/creation/good-practices.md
```diff
@@ -69,6 +69,8 @@ menuTitle: Good practices
 - A merchant is likely to have a shop running on a different shop than yours. In case a module is using PHP extensions not installed by default by PHP,
 add a preliminary check before using them (I.e with `extension_loaded`). This prevents fatal errors to be thrown on shops on which these extensions aren’t enabled.
 
+- About files, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scallable applications, and also to keep data if needed in case of you uninstall the module.
```
kpodemski:
```suggestion
- About files, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scalable applications, and also to keep data if needed in case of you uninstall the module.
```

modules/creation/good-practices.md
```diff
@@ -69,6 +69,8 @@ menuTitle: Good practices
 - A merchant is likely to have a shop running on a different shop than yours. In case a module is using PHP extensions not installed by default by PHP,
 add a preliminary check before using them (I.e with `extension_loaded`). This prevents fatal errors to be thrown on shops on which these extensions aren’t enabled.
 
+- About files, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scalable applications, and also to keep data if needed in case of you uninstall the module.
+
```
PululuK: to avoid confusion, I think it might be interesting to specify in which context this recommendation is valid, because in some cases storing data outside the module maybe a bad idea too

kpodemski: That's a good point.

modules/creation/good-practices.md
```diff
@@ -69,6 +69,8 @@ menuTitle: Good practices
 - A merchant is likely to have a shop running on a different shop than yours. In case a module is using PHP extensions not installed by default by PHP,
 add a preliminary check before using them (I.e with `extension_loaded`). This prevents fatal errors to be thrown on shops on which these extensions aren’t enabled.
 
+- About files, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scalable applications, and also to keep data if needed in case of you uninstall the module.
```
PierreRambaud:
```suggestion
- About files, if you plan to save customer data generated by your module, we also recommend you to save them inside a `/var/modules/YOUR_MODULE/` directory. Putting them outside the `/modules` directory will help for scalable applications, and also to keep data if needed in case of you uninstall the module.
```

PierreRambaud: @kpodemski @PululuK is it better?

eternoendless:
```suggestion
- If you need to store static files: we recommend putting temporary or cache files in prestashop's `/var/cache/<env>/modules/YOUR_MODULE/` directory, and static or shared files in prestashop's  `/var/modules/YOUR_MODULE/` directory. Writing in the `/modules` directory is not recommended, because doing so will make it harder to use your module in distributed environments. In addition, by storing files outside your module's directory, they are kept even if the module is uninstalled.
```

atomiix (approved):

atomiix:
Thanks @PierreRambaud

eternoendless:
Needs to be backported into 1.7.x

matks:
This doc change is only relevant for `8.x` branch

PR #1137, opened 2021-09-08
Update URL of author inside the author configuration

PrestaEdit:
| Questions     | Answers
| Branch?       | 1.7.x
| Description?  | As author URL is excluded from PrestaShop Addons rules, mention it.

### DIFF
--- a/themes/getting-started/theme-yml.md
+++ b/themes/getting-started/theme-yml.md
@@ -21,7 +21,7 @@ Users will be able to choose the layout for each page from the theme's settings 
   author:
     name: "John Doe"
     email: "pub@prestashop.com"
-    url: "https://www.prestashop.com"
+    url: "https://www.prestashop.com" # Avoided by PrestaShop Addons, don't add it if you want to distribute the theme on Addons Platform
   meta:
     compatibility:
         from: 1.7.0.0
--- a/themes/getting-started/theme-yml.md
+++ b/themes/getting-started/theme-yml.md
@@ -21,7 +21,7 @@ Users will be able to choose the layout for each page from the theme's settings 
   author:
     name: "John Doe"
     email: "pub@prestashop.com"
-    url: "https://www.your-url.com"
+    url: "https://www.your-url.com" # Be aware that some marketplaces don't allow this parameter
   meta:
     compatibility:
         from: 1.7.0.0
--- a/themes/getting-started/theme-yml.md
+++ b/themes/getting-started/theme-yml.md
@@ -21,7 +21,7 @@ Users will be able to choose the layout for each page from the theme's settings 
   author:
     name: "John Doe"
     email: "pub@prestashop.com"
-    url: "https://www.your-url.com" # Be aware that some marketplaces don't allow this parameter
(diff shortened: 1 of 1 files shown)


Progi1984 (commented):
I'm not a big fan of this PR.

Why ? We are currently in split of Company and Project. So we try to remove specific informations about Addons.

PrestaEdit:
I know. But I need to make this PR in order to remember be this situation.
We could maybe remove the "theme_key" and add this specific part (theme_key and add notice about URL) on the "Distribute on Addons".

matks:
The "Distribute on Addons" section is supposed to move in another website in the future, so that knowledge is not lost. It will be somewhere near Addons naturally.

PrestaEdit:
_Supposed, future, maybe, ..._

So, it's not done, yet.
So, we remove this now or we could add this kind of informations as themes developpers facing this problem on market validation right now, not in the future. :)

themes/getting-started/theme-yml.md
```diff
@@ -21,7 +21,7 @@ Users will be able to choose the layout for each page from the theme's settings
   author:
     name: "John Doe"
     email: "pub@prestashop.com"
-    url: "https://www.prestashop.com"
+    url: "https://www.prestashop.com" # Avoided by PrestaShop Addons, don't add it if you want to distribute the theme on Addons Platform
```
PierreRambaud:
```suggestion
    url: "https://www.prestashop-project.org/"
```

PierreRambaud:
Let's change the URL by `https://www.prestashop-project.org/` and we don't care about what Addons wants or not. An author should be defined by a name, an email and an URL.

themes/getting-started/theme-yml.md
matks: Good idea IMO 👍

eternoendless: Even better:
```suggestion
    url: "https://www.your-url.com" # Be aware that some marketplaces don't allow this parameter
```

PrestaEdit:
Hi all,

I've removed the notice about Addons, and so the title of the PR but note sure that it's really needed, so, to merge this as it.

Progi1984 (approved):

themes/getting-started/theme-yml.md
```diff
@@ -21,7 +21,7 @@ Users will be able to choose the layout for each page from the theme's settings
   author:
     name: "John Doe"
     email: "pub@prestashop.com"
-    url: "https://www.prestashop.com"
+    url: "https://www.your-url.com"
```
atomiix: Could you add the comment as stated by @eternoendless in [this comment](https://github.com/PrestaShop/docs/pull/1137#discussion_r727312770)?

atomiix (approved):

matthieu-rolland (approved):

matthieu-rolland:
thank you @PrestaEdit 👍

eternoendless:
This will need to be backported into 8.x

matks:
> This will need to be backported into 8.x

`8.x` branch content is updated

PR #1154, opened 2021-09-24
Fix Position location

TheDreamCatcher:
| Questions     | Answers
| Branch?       | 1.7.x
| Description?  | The position page is inside the Design one, not inside the Modules part

### DIFF
--- a/modules/creation/displaying-content-in-front-office.md
+++ b/modules/creation/displaying-content-in-front-office.md
@@ -99,7 +99,7 @@ which generates the correct `<script>` tag.
 Save your file, and already you can hook your module's template into the
 theme, move it around and transplant it (even though there is not
 template file for the moment): go to the "Positions" page from the
-"Modules" menu in the back office, then click on the "Transplant a
+"Design" menu in the back office, then click on the "Transplant a
 module" button (top right of the page).
 
 In the transplantation form:


SharakPL:
Yeah, that has always confused me. If it's about modules, then why is it suddenly in Design section and not in Modules section, where it was in Prestashop 1.6?

kpodemski:
Because hooks that you can sort are more likely to be used on the front office (Design of the store)

SharakPL:
1. You can't sort hooks. To do that you need to change their placement in .tpl files. You can sort modules inside specific hook, so again: it's about modules more than about design.
2. Both front office and back office use the same page for changing **modules' positions** 😉 These last 2 words are self-explanatory, Design -> Positions is not that obvious.

At least it should be changed to _Design -> Modules' position_ or _Design -> Hooks_

matks (approved):

matks:
Thank you @TheDreamCatcher 

@SharakPL if you want to challenge the position in the menu, you will have to talk to @PrestaShop/product-team on https://github.com/prestashop/prestashop 😉

SharakPL:
Thanks @matks

matks:
PR to update 8 branch https://github.com/PrestaShop/docs/pull/1158

PR #1164, opened 2021-10-02
Update setting-up-your-local-environment.md

vahidalvandi:
Git clone fast . 
Remove history of commit

| Questions     | Answers
| Branch?       | 1.7.x / 8.x

### DIFF
--- a/themes/getting-started/setting-up-your-local-environment.md
+++ b/themes/getting-started/setting-up-your-local-environment.md
@@ -20,7 +20,7 @@ Open a command line on your (empty) working directory, then:
 
 1. Clone PrestaShop
     ```bash
-    git clone https://github.com/PrestaShop/PrestaShop.git
+    git clone --depth 1 https://github.com/PrestaShop/PrestaShop.git
     ```
     
 2. Install dependencies


matks (changes requested):
Hi, I think depending on one's goal, some people might want the history and some people might dont want it.

We cannot recommend the `--depth 1` unless we explain carefully what it will do. There are beginner developers who read this documentation 😉 we need to make sure they understand what this line will do.

vahidalvandi:
this is very simple command  `--depth 1`  ignore all history commit . for explanation , when you send commit to github all changes store and after clone repo all history change in all time not need . this with ignore get all history 74,210 commits !!!!

PierreRambaud:
> this is very simple command `--depth 1` ignore all history commit . for explanation , when you send commit to github all changes store and after clone repo all history change in all time not need . this with ignore get all history 74,210 commits !!!!

I still see no reason to recommend people to use this parameter.

matks:
I guess it's useful to ignore history if the aim is to create a new shop from the sources. But this doc is for people who want to build the project locally and submit contributions, not for people who create new shops. The standard creation process is to use the ready-for-install ZIP 🤔 .

I'm sorry but I don't think we should add this in the devdocs, so I close the PR.

PR #1186, opened 2021-10-13
Fix wrong directory in CI when in pull request

eternoendless:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Fixed the directory used in CI when building for pull requests. Only push was working because github.ref is the name of the branch being buiolt, not the destination branch.
| Fixed ticket? | N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,6 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    env:
-      TARGET_BRANCH: ${{ github.ref }}
-
     steps:
       - name: Checkout Hugo Site
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


matks (commented):
Bash script is not the most readable we can do ^^

Would it be overkill to put this bash script in a file in `.github` and a test to it? This stuff is fragile and is easy to break if we modify it later

PierreRambaud (approved):

eternoendless:
This is not a bash script @matks although it does have some bash in it. I don't understand what you have in mind actually.

> This stuff is fragile and is easy to break if we modify it later

Let's try not to modify it then 😛

matks:
> This is not a bash script @matks although it does have some bash in it. I don't understand what you have in mind actually.

I was meaning these parts:
```
echo $TARGET_BRANCH | perl -pe 's|^([\d.]+)(?:\.x)?$|\1|'
```

```
          VERSION_FORMAT='^([0-9]+\.)*[0-9]+$'
          if [[ ! $VERSION =~ $VERSION_FORMAT ]]; then
            echo "::error::Version '$VERSION' is not a valid version!"
            exit 1
          else
            echo "Version: '$VERSION'"
          fi
```

matks (approved):

eternoendless:
Merging to unblock CI, but discussion can continue

PR #1187, opened 2021-10-13
Upgrade files are now in autoupgrade module

PierreRambaud:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Files are now in the autoupgrade modules and are going to be removed from the core

### DIFF
--- a/basics/keeping-up-to-date/upgrade.md
+++ b/basics/keeping-up-to-date/upgrade.md
@@ -128,14 +128,14 @@ Some web hosting providers gives you two user accounts to access your database. 
 All the changes to apply have been defined in the `install` folder,
 running them can be done with a specific PHP script.
 
-When you’re ready, run the file `install/upgrade/upgrade.php`.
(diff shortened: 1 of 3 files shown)


ghost (approved):

matks (approved):

matks:
Thank you @PierreRambaud

matks:
@PierreRambaud Should we backport this PR or not? 🤔  does it make sense for 1.7 doc ?

PierreRambaud:
> @PierreRambaud Should we backport this PR or not? thinking does it make sense for 1.7 doc ?

I think yes :thinking: even if we remove scripts only for the 8.0, we will now handle SQL and php scripts in the next version of the autoupgrade module

matks:
We will have to make sure people understand that
- in PS < 8.0.0 there is SQL scripts inside the Core, for all 1.7 versions
- the autoupgrade module < 5.0.0 use these scripts inside the Core
- the autoupgrade module 5.0.0 use its own scripts

PR #1194, opened 2021-10-20
Update broken 404 links to github points 1.7.6.x -> 1.7.8.0

SebSept:
links to GitHub points to 1.7.6.x which doesn't exist anymore.

| Questions     | Answers
| Branch?       | 1.7.x
| Description?  | 404 not found - fix links that points to 1.7.8.x branch instead of 1.7.6.x which is removed from GH
| Fixed ticket? | none

### DIFF
--- a/modules/concepts/controllers/front-controllers.md
+++ b/modules/concepts/controllers/front-controllers.md
@@ -43,7 +43,7 @@ automatically.
 
 ## Available properties
 
-The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.x/classes/controller/Controller.php).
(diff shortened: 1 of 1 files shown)


matks (changes requested):
Hello !

Links to 1.7.6.x are broken because we deleted the 1.7.6.x branch. But I am not sure 1.7.8.x is a good replacement. One day we will delete 1.7.8.x branch too.

I suggest to rather use `tags` link:
https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x
⬇️ 
https://github.com/PrestaShop/PrestaShop/blob/1.7.6.9

SebSept:
or maybe 1.7.8.0 ?
https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/FrontController.php

matks:
> or maybe 1.7.8.0 ? https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/FrontController.php

Could be, but as 1.7.6.x was used when this documentation was written, I am 100% sure the doc describes 1.7.6 situation well.

We can use 1.7.8.0 if we are 100% sure it properly illustrates this documentation section 😉

SebSept:
Right, I'll check that in a few days.

Progi1984:
@SebSept Ping us after check.

SebSept:
Nothing serious and related to the `module` property has changed, since 2016 ... at least. 
PrestaShop version was 1.7.0.0 release candidate. Since mid 2020 (1.7.6 release), no changes except formating and licenses texts.
I also compared Controller.php (which hold the  `context` property) with a version before 1.7.6 and there is no changes related to `context`.
ping @Progi1984 / @matks

matks (approved):

matks (approved):

matks:
I updated the PR to use the 1.7.8.0 tag instead of 1.7.8.x branch. 1.7.8.x branch will be deleted in the future, 1.7.8.0 will not.

matks:
Thank you @SebSept

matks:
@SebSept Could you submit another PR to backport the changes to 8.x branch?

SebSept:
Thanks @matks, done for branch 8 -> #1208

PR #1196, opened 2021-10-22
Updating context of where code goes

Wrenbjor:
It's not clear that we are still working with the mymodule.php file because the code block starts with a new <?php block which tricked me into thinking that we were in a new file. I added a line to make sure it's known to work in the mymodule.php file.

| Questions     | Answers
| Branch?       | 1.7.x


matks (approved):
Thank you this is a nice improvement that makes documentation less confusing 👍

PululuK (approved):

PululuK:
Thanks @Wrenbjor

matks:
@Wrenbjor Could you submit another PR to update the 8.x branch with your improvement? 😉

Wrenbjor:
@matks please see #1198

PR #1208, opened 2021-11-10
Fixed links to ModuleFrontController.php

SebSept:
same as #1194

| Questions     | Answers
| Branch?       | 8.x
| Description?  | fix links to ModuleFrontController.php
| Fixed ticket? | none

### DIFF
--- a/modules/concepts/controllers/front-controllers.md
+++ b/modules/concepts/controllers/front-controllers.md
@@ -41,7 +41,7 @@ automatically.
 
 ## Available properties
 
-The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/Controller.php).
+The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/Controller.php).
 They provide access to the environment in which they run.
 
 * `$this->module` is the instance of the module responsible of the controller.


modules/concepts/controllers/front-controllers.md
```diff
@@ -41,7 +41,7 @@ automatically.
 
 ## Available properties
 
-The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/1.7.6.x/classes/controller/Controller.php).
+The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/1.7.8.0/classes/controller/Controller.php).
```
kpodemski, line 44:
```suggestion
The controllers added in a module extend [`ModuleFrontController`](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/ModuleFrontController.php), itself extending [`FrontController`](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/FrontController.php) & [`Controller`](https://github.com/PrestaShop/PrestaShop/blob/develop/classes/controller/Controller.php).
```

`develop` is the next major (v8), maybe it would be better to link these URLs to this branch? :)

matks: But develop will move 😅  so in future maybe these links will break.

The best would be to have a 8.0.0 tag but it's not there yet 🤷

kpodemski: @matks you're absolutely right, maybe we should create it? ;)

matks: But we cant, a tag has to be linked to a release version :(

kpodemski: Right, I thought we could force the commit related to the tag later on, but other things block us.

In that case, I'm not sure what would be the best way 🤔 it seems that linking to 1.7.8.x branch is the safest option? (although we can expect to see this kind of PR in the future, again)

matks: Maybe we have to use 1.7.8.x for now and, when we release 1.7.8.0, we change it.

PierreRambaud: If it's valid for 1.7.8.x no need to wait for 8.0, and the PR needs to be backported in the 1.7 doc :thinking:

matks (approved):

matks:
Thank you @SebSept

PR #1224, opened 2021-11-18
Update contribute_using_localhost.md

rickygzz:
Suggest to add PHP extensions (gd and intl) in PHP.ini which are needed in order for `composer install` to work.

| Questions     | Answers
| Branch?       | 1.7.x / 8.x
| Description?  | Suggest to add PHP extensions (gd and intl) in PHP.ini which are needed in order for `composer install` to work.
| Fixed ticket? |

### DIFF
--- a/contribute/contribute-pull-requests/contribute_using_localhost.md
+++ b/contribute/contribute-pull-requests/contribute_using_localhost.md
@@ -80,6 +80,8 @@ cd /path/to/prestashop
 composer install
 ```
 
+You may need to edit `PHP.ini` to add or uncomment PHP extensions `extension=gd` and `extension=intl`.
+
 ### Compile assets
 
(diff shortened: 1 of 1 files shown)


PierreRambaud (changes requested):
Not sure it's a good idea.
Modern systems are automatically include extensions, editing the php.ini file directly for this kind of stuff isn't a good idea:
image

The scan is done by [with-config-file-scan-dir](https://www.php.net/manual/en/configuration.file.php) at compile time
And also because there are a lot of mandatory extensions:
image

matks:
Hello @rickygzz I agree with @PierreRambaud : I think this knowledge is too specific to environment and might confuse users. I think the devdocs should focus on PrestaShop and no attempt to explain the wide topic of environments 😉 .

I close this PR. Thank you for your contribution.

PR #1238, opened 2021-12-01
Remove notice in development/architecture/introduction

mohamedaitbella:
remove 
notice note 
"While in this diagram we have placed Modules on the Back-end side, they can actually have an impact on Front-end as well."

because the  diagram did not placed Modules on the Back-end side

| Questions     | Answers
| Branch?       | 1.7.x
| Description?  | remove notice note 

### DIFF
--- a/development/architecture/introduction.md
+++ b/development/architecture/introduction.md
@@ -58,10 +58,6 @@ While Web services can output XML or JSON data, the BO API is JSON-only.
 
 Finally, **Modules** are independent, optional packages that can extend and customize PrestaShop in many ways. They interact with the Core either by _hooking_ into extension points which are placed throughout the code, or by _replacing_ core components with their own.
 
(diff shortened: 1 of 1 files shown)


matks (changes requested):
Hello @mohamedaitbella I think there is, yes, an issue in https://devdocs.prestashop.com/1.7/development/architecture/introduction/ but not the one.

I think this note aimed to remember that modules can modify the software behavior both on backend (on server) and frontend (in browser) 😄  .

It's a usual ambiguity between back-office / backend and front-office / frontend.

So I dont think we should remove this, but instead we should IMPROVE this. Wdyt ?

mohamedaitbella:
yes a change can  remove the ambiguity : 
somthing like this :
 in this diagram Modules takes its place  between Back-end side and Front-end because they can have an impact on them both

PR #1243, opened 2021-12-09
Removing 'Selling your modules' from docs

### LINKED ISSUE #1108: Removing "Selling your modules" from docs
With the separation of the company and the project, we need to remove docs about Addons: https://devdocs.prestashop.com/1.7/modules/sell/

Progi1984:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Removing 'Selling your modules' from docs
| Fixed ticket? | Fixes #1108


matks:
Maybe we should wait for https://validator.prestashop.com/ to be deployed with this info before removing it? If we remove it from devdocs now, people that need this info will be left alone in the dark (nobody likes to be left alone in the dark 😄 ).

And when validator.prestashop.com provides the doc, maybe we should add a temporary redirect, to help people who bookmarked this link to find the new location.

For years this doc has lived here. It's not the right place anymore, thanks to the split between PrestaShop company and PrestaShop project but we need to help making the transition smooth :) confusing the people who have been reading this doc for years is not a good thing to achieve. I think we can provide the little help for people to understand where things are going, what will remain on devdocs.prestashop.com, what will go somewhere else. Also we have still not announced publicly and clearly the split 😅  we talked about it in events, we talked about in on YouTube but we still have not published a blog post that explains exactly what will happen (should be done soon!).

Progi1984:
@matks The information will be always in 1.7 branch, which is the default one. The removal is in 8.x branch. So I think we can merge it.

matks (approved):

matks:
> The information will be always in 1.7 branch, which is the default one. The removal is in 8.x branch. So I think we can merge it.

Good point, I approve and merge.

That being said I think we can also remove it from 1.7.x branch 😄  just not today

PR #1247, opened 2021-12-10
Port PR #1098

matks:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Port PR #1098 into FAQ section
| Fixed ticket? |

### DIFF
--- a/faq/theme.md
+++ b/faq/theme.md
@@ -1,0 +1,15 @@
+---
+title: Theme FAQ
+---
+
+# Theme FAQ
+
+**Q:** Why new-theme directory does not have a package.json?
+
+**A:** If you use PrestaShop's production ZIP release instead of the sources, new-theme directory will not have a package.json. `npm run build` will then fail.
+
+```
+Module build failed: Error: Couldn't find preset "env" relative to directory
+```
+
+You can solve this by changing to a release for development or by copy the necessary json file to your new-theme folder, navigate there and run `npm install`.


faq/theme.md
```diff
@@ -0,0 +1,15 @@
...
+
+**Q:** Why new-theme directory does not have a package.json?
+
+**A:** If you use PrestaShop's production ZIP release instead of the sources, new-theme directory will not have a package.json. `npm run build` will then fail.
+
+```
+Module build failed: Error: Couldn't find preset "env" relative to directory
+```
+
+You can solve this by changing to a release for development or by copy the necessary json file to your new-theme folder, navigate there and run `npm install`.
```
PierreRambaud, line 15: IMHO, the production build shouldn't be tweaked. But the source must be used instead

matks: You're right, the best scenario is to use sources. But I discussed with multiple agency developers who do not 😢  : they download production-ready ZIP and copy the files they need from github.

So what do we do: we explain the best usecase that maybe most people dont follow, or we add a FAQ item for that we know is a real life situation 🗃️

PierreRambaud:
```suggestion
PLEASE DON'T CHANGE THE PRODUCTION ZIP, USE SOURCE AND CREATE A RELEASE IF NEEDED!
```

matks: @PierreRambaud I think I can guess what you mean 😄

PR #1178, opened 2021-10-10
Document configuration console command

tswfi:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | develop
| Description?  | Add documentation about configuration console command
| Fixed ticket? | 

This should only be merged after https://github.com/PrestaShop/PrestaShop/pull/26183 is merged and the minver number should be double checked at that moment also.

PR #1252, opened 2021-12-15
Document configuration console command

tswfi:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Add documentation about configuration console command
| Fixed ticket? | 

Document the new configuration console command which was merged here: https://github.com/PrestaShop/PrestaShop/pull/26183 

Replaces #1178 as rebase took way too long :)

### DIFF
--- a/development/components/configuration/_index.md
+++ b/development/components/configuration/_index.md
@@ -143,3 +143,23 @@ $configuration->set(
     ShopConstraint::shopId(12)
 );
 ```
+
+## Console command
+{{< minver v="8.0" title="true" >}}
+
+You can also use console command to get, set and remove config values from command line.
+
+### Set smtp settings to use mailcatcher for developing
+
+```bash
+$ bin/console prestashop:config set PS_MAIL_SERVER --value "my-mailcatcher-container-name"
+PS_MAIL_SERVER="my-mailcatcher-container-name"
+$ bin/console prestashop:config set PS_MAIL_SMTP_PORT --value 1025
+PS_MAIL_SMTP_PORT="1025"
+```
+
+To see all the available actions use:
+
+```bash
+$ bin/console prestashop:config --help
+```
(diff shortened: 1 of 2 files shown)


matks (commented):
Looks good to me. We could also add it here: https://devdocs.prestashop.com/1.7/development/components/console/

What do you think ? We add the informations to the 2 locations? Only one?

tswfi:
hmm, that page looks like something that should be automated.. 

github action that checks for merges to core
runs --help for all prestashop commands (can these be identified somehow?)
checks the result against the current documentation
if there are differences create a new pr to documentation "Changes to commands documentation found, requires human to review"

It should probably also understand the situation where the human review was not approved so it should check for documentation pr:s for that command and if there are closed pr:s that have not been merged not open a new one..

or something :)

matks:
> hmm, that page looks like something that should be automated..
> 
> github action that checks for merges to core runs --help for all prestashop commands (can these be identified somehow?) checks the result against the current documentation if there are differences create a new pr to documentation "Changes to commands documentation found, requires human to review"

Maybe this would be overkill 😅 ? I'm worried if we build this, we would spend more time fixing the CI and edge cases than the time we would need to write the doc manually

https://xkcd.com/974/

tswfi:
Maybe :D

https://xkcd.com/1205/

matthieu-rolland (commented):
I agree that it would make sense to have it in the console page too 👍

tswfi:
Yup, now I had some free time to set up a local devdocs site so that I can see what I'm doing :D

tswfi:
Still think it might be worth it to write a script that could build the "base" for each cli command :)

PierreRambaud (approved):

Progi1984 (approved):

Progi1984:
Thanks @tswfi

PR #1254, opened 2021-12-16
Refer to nodejs compatibility table and add node 16

NeOMakinG:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | We need to support Node16 which is the latest LTS and kick the old nodejs version in getting started

⚠️ we need to merge this https://github.com/PrestaShop/PrestaShop/pull/27006 before updating the doc, because we're not fully compatible with node 16 yet


matthieu-rolland (approved):

PierreRambaud (approved):

PierreRambaud:
Thank you @NeOMakinG

PR #1267, opened 2022-01-11
Changed database structure version number to the latest one

anttivu:
| Questions     | Answers
| Branch       | 1.7.8
| Description | Changed database structure version number to the latest one. However, this is not a definitive solution to the problem. There should be a link that points allways to the current /structure.md -file, so the link doesn't need to be changed every time a new version is published.?

### DIFF
--- a/development/database/structure.md
+++ b/development/database/structure.md
@@ -9,7 +9,7 @@ weight: 4
 
 ### Global definition
 
-The database structure of PrestaShop can be found in `install/data/db_structure.sql` ([1.7.3.x releases example](https://github.com/PrestaShop/PrestaShop/blob/1.7.3.x/install-dev/data/db_structure.sql)).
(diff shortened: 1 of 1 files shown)


PululuK:
@anttivu need a generic name something like :
https://github.com/PrestaShop/PrestaShop/blob/your_prestashop_version/install-dev/data/db_structure.sql

Or juste use develop branch ?

anttivu:
Yes generic name pointing allways to the latest version would do. Using develop branch would also do. Anyway link pointing to an existing and the latest version would be great.

PierreRambaud (approved):

PierreRambaud:
I agree with @anttivu it should target the latest available 1.7.x branch, and should never target the develop branch.
The reason is, imagine in 9.x you change the db_structure file name, or move it to another directory, it will break all previous docs.

matks (approved):

matks:
Agreed with @PierreRambaud 

Thank you @anttivu

PR #1272, opened 2022-01-24
Mention the need to use namespace import of code example is used outside of initial namespace

WebHelpersPau:
| Questions     | Answers
| Branch?       | 1.7.x
| Description?  | Missing inclusions on example 
| Fixed ticket? | N/A

### DIFF
--- a/development/architecture/migration-guide/forms/CRUD-forms.md
+++ b/development/architecture/migration-guide/forms/CRUD-forms.md
@@ -53,7 +53,6 @@ In the example below, you can see a `ContactFormDataProvider` that queries the d
 namespace PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider;
 
 use Contact;
-use PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider\FormDataProviderInterface;
 
 final class ContactFormDataProvider implements FormDataProviderInterface
 {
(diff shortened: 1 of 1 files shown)


development/architecture/migration-guide/forms/CRUD-forms.md
```diff
@@ -53,6 +53,7 @@ In the example below, you can see a `ContactFormDataProvider` that queries the d
 namespace PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider;
 
 use Contact;
+use PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider\FormDataProviderInterface;
```
matks: I dont think this is necessary. Classes and interfaces that are in the same namespace `PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider` are automatically found if I am not wrong?

WebHelpersPau: Not on mine, I've been facing it in a module creation...

matks: Hello @WebHelpersPau and apologies for late answer.

I think if you use the above code in the namespace `PrestaShop\PrestaShop\Core\Form\IdentifiableObject\DataProvider` you do not need this `use`.

But you used this code in a module creation, which means the namespace was not the same which means the `use` was needed.

So in order to address both situations I will add a mention in the doc ;) so we can have both solutions that fit well with the need.

matks: See https://github.com/PrestaShop/docs/pull/1272/commits/a9f795465169a0bf318bfccd4a0e9be278ad80c0

matks (approved):

matks:
Thank you @WebHelpersPau

matks:
Backported to 8.x https://github.com/PrestaShop/docs/pull/1309

PR #1278, opened 2022-01-25
Please clarify the correct way to autoload Commands added by modules

ioweb-gr:
I'm adding the steps I had to do to actually make the command accessible and working on our Prestashop 1.7.5.2 installation as I could find no other documentation on the proper procedure apart from a sample module that can be installed via a custom repository. Unfortunately that way of installing wasn't possible in our case and we had to work with an existing legacy module that needed to append a new console command for handling extensively long cron tasks. 

This is documentation purely based on experience on our installation and maybe it will help others looking to do the same. It doesn't mean it's the proper way, it's the only way I could find by experience to make things work. 

If there's a suggested way of doing it, please scratch this out and write it so that we can use everything correctly in the recommended way.

| Questions     | Answers
| Branch?       | 1.7.x / 8.x
| Description?  | The documentation is not very intuitive and I struggled for hours to make it work. That's why I enhanced by describing my experience with this area of Prestashop

### DIFF
--- a/faq/modules.md
+++ b/faq/modules.md
@@ -1,0 +1,59 @@
+---
+title: Modules FAQ
+---
+
+# Modules FAQ
+
+[module-commands]: {{< ref "/1.7/modules/concepts/commands" >}}
+
+**Q:** How do i add a command to an existing module, with autoloading?
+
+**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions are assuming the module is installed via composer. In case you follow the same procedure but wish to add the command in your already existing module in `modules/<module_directory>` then the procedure requires an extra step. 
+
+For example with a structure like this:
+
+```tree
+modules/module_directory/
+├── config
+│   └── services.yml
+├── config.xml
+├── module_directory.php
+├── src
+│   └── Command
+│       └── SyncCommand.php
+```
+
+You would also need to namespace your src directory in `composer.json` located in the base directory of your project, by appending the following:
+
+```
+    "autoload": {
+        "psr-4": {
+            "PrestaShop\\PrestaShop\\": "src/",
+            "PrestaShopBundle\\": "src/PrestaShopBundle/",
+            "Vendor\\Module\\": "modules/module_directory/src/"
+        },
+        "classmap": [
+            "app/AppKernel.php",
+            "app/AppCache.php"
+        ]
+    },
+```
+
+And follow through using the correct namespaces throught your module commands. 
+
+**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_` you also need to require the file `config/config.inc.php` 
+
+e.g. 
+
+```php
+$basePath = realpath(__DIR__ . '/../../../../');
+require_once $basePath . '/config/config.inc.php';
+```
+
+**Loading Module Classes**: Commands by default won't have module classes loaded. You need to explicitly require the modules before you can use their classes
+
+e.g. 
+
+```php
+require_once _PS_MODULE_DIR_ . '/<module_directory>/<module_directory>.php';
+```        
--- a/faq/modules.md
+++ b/faq/modules.md
@@ -6,11 +6,11 @@ title: Modules FAQ
 
 [module-commands]: {{< ref "/1.7/modules/concepts/commands" >}}
 
-**Q:** How do i add a command to an existing module, with autoloading?
+**Q:** How do I add a command to an existing module with autoloading?
 
-**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions are assuming the module is installed via composer. In case you follow the same procedure but wish to add the command in your already existing module in `modules/<module_directory>` then the procedure requires an extra step. 
+**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions assume the module is installed via Composer. If you follow the same procedure but wish to add the command in your existing module in `modules/<module_directory>`, then the procedure requires an extra step. 
 
-For example with a structure like this:
+For example, with a structure like this:
 
 ```tree
 modules/module_directory/
@@ -23,7 +23,7 @@ modules/module_directory/
 │       └── SyncCommand.php
 ```
 
-You would also need to namespace your src directory in `composer.json` located in the base directory of your project, by appending the following:
+You would also need to namespace your `src` directory in `composer.json` located in the base directory of your project, by appending the following:
 
 ```
     "autoload": {
@@ -39,9 +39,9 @@ You would also need to namespace your src directory in `composer.json` located i
     },
 ```
 
-And follow through using the correct namespaces throught your module commands. 
+And follow through using the correct namespaces throughout your module commands. 
 
-**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_` you also need to require the file `config/config.inc.php` 
+**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_`, you also need to require the file `config/config.inc.php`. 
 
 e.g. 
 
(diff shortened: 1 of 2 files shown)


modules/concepts/commands.md
```diff
@@ -77,6 +77,66 @@ services:
...
+```
+
+### Loading Module Classes 
+
+Commands by default won't have module classes loaded. You need to explicitly require the modules before you can use their classes
+
+e.g. 
+
+```php
+require_once $basePath . '/modules/<module_directory>/<module_directory>.php';
```
kpodemski: @ioweb-gr I would go with `_PS_MODULE_DIR_` instead of `$basePath`

ioweb-gr: I can see your point. 

I opted to do it this way because in some commands, the constants may not be declared. 

As mentioned in the PR, if you don't require config/config.inc.php the constants are not declared in the command and they're not accessible. For some cases you may not have the need to work with prestashop core classes but instead want to do something entirely irrelevant or agnostic to the prestashop core. So I wouldn't require config/config.inc.php in this case 

Perhaps we could write both ways but it is a bit overkill isn't it?

modules/concepts/commands.md
```diff
@@ -77,6 +77,66 @@ services:
...
+├── config
+│   └── services.yml
+├── config.xml
+├── module_directory.php
+├── src
+│   └── Command
+│       └── SyncCommand.php
+```
+
+You would also need to namespace your src directory in `docroot/composer.json`
```
matks: I'm not sure to fully understand what docroot is

modules/concepts/commands.md
```diff
@@ -77,6 +77,66 @@ services:
...
+
+By appending the following 
+
+```
+    "autoload": {
```
matks: Are you talking about the root composer.json file ? https://github.com/PrestaShop/PrestaShop/blob/develop/composer.json

matks (commented):
First, thank you for your contribution.

If I understand this PR, what you mean is that _it is not possible today to add Commands inside a Module_ because PrestaShop does not "find" it, right?

But that actually sounds like a bug to me 😅  ?

modules/concepts/commands.md
ioweb-gr: docroot is a term from apache to refer to the document root where the website resides. It's the directory where each website resides in classic web apps.

modules/concepts/commands.md
ioweb-gr: Yes

ioweb-gr: Also I don't think it's actually a bug, it's more like a symfony command is declared outside of the prestashop core bundle so it's normal that the namespace is unknown to external modules unless you declare it.

I guess creating a composer plugin that extends the composer autoloader to find those namespaces when composer is dumping the autoloader would automate this but I don't want to refer to missing functionality as a bug

modules/concepts/commands.md
matks: We might need to use another word if we want this documentation to be as easy as possible for all type of readers ;)

modules/concepts/commands.md
matks: Thank you for explanation.
You are right this is a missing functionality rather than a bug, but nonetheless this is something PrestaShop should be able to do one day.

As this piece of doc you have written is a nice workaround the current limitation I'm not sure we should put it in the Command page. Maybe in the FAQ section it would be better? https://devdocs.prestashop.com/1.7/faq/

What you have written has value 👍  but it's a how-to that aim to help others looking to do it. It's away you could find by experience to make things work but the standard doc should talk about 'standard and recommended ways' I guess

wdyt @PrestaShop/prestashop-maintainers

kpodemski: Thank you, @ioweb-gr - this can be useful for some.

In my opinion, it's more like a workaround than a preferred way of doing things, and I think it would be better to have it in the FAQ section. I'll leave the final decision to @thomasnares

thomasnares: Yes it should be definitely in FAQ section, can you move it to FAQ @ioweb-gr ?
Thanks 👍

thomasnares:
Proposed the PR on @ioweb-gr 's repo : https://github.com/ioweb-gr/docs/pull/1

thomasnares (approved):

thomasnares:
@kpodemski imo, this one can be merged

faq/modules.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Modules FAQ
+---
+
+# Modules FAQ
+
+[module-commands]: {{< ref "/1.7/modules/concepts/commands" >}}
+
+**Q:** How do i add a command to an existing module, with autoloading?
```
kpodemski:
```suggestion
**Q:** How do I add a command to an existing module with autoloading?
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+
+**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions are assuming the module is installed via composer. In case you follow the same procedure but wish to add the command in your already existing module in `modules/<module_directory>` then the procedure requires an extra step.
```
kpodemski:
```suggestion
**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions assume the module is installed via Composer. If you follow the same procedure but wish to add the command in your existing module in `modules/<module_directory>`, then the procedure requires an extra step. 
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Modules FAQ
+
+[module-commands]: {{< ref "/1.7/modules/concepts/commands" >}}
+
+**Q:** How do i add a command to an existing module, with autoloading?
+
+**A:** [In the modules section of the documentation, about module commands][module-commands], the instructions are assuming the module is installed via composer. In case you follow the same procedure but wish to add the command in your already existing module in `modules/<module_directory>` then the procedure requires an extra step. 
+
+For example with a structure like this:
```
kpodemski:
```suggestion
For example, with a structure like this:
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+├── config
+│   └── services.yml
+├── config.xml
+├── module_directory.php
+├── src
+│   └── Command
+│       └── SyncCommand.php
+```
+
+You would also need to namespace your src directory in `composer.json` located in the base directory of your project, by appending the following:
```
kpodemski:
```suggestion
You would also need to namespace your `src` directory in `composer.json` located in the base directory of your project, by appending the following:
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+            "Vendor\\Module\\": "modules/module_directory/src/"
+        },
+        "classmap": [
+            "app/AppKernel.php",
+            "app/AppCache.php"
+        ]
+    },
+```
+
+And follow through using the correct namespaces throught your module commands.
```
kpodemski:
```suggestion
And follow through using the correct namespaces throughout your module commands. 
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+        "classmap": [
+            "app/AppKernel.php",
+            "app/AppCache.php"
+        ]
+    },
+```
+
+And follow through using the correct namespaces throught your module commands. 
+
+**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_` you also need to require the file `config/config.inc.php`
```
kpodemski:
```suggestion
**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_`, you also need to require the file `config/config.inc.php`. 
```

faq/modules.md
```diff
@@ -0,0 +1,59 @@
...
+**Predefined Constants**: If you would like to use existing PS constants like `_DB_PREFIX_` you also need to require the file `config/config.inc.php` 
+
+e.g. 
+
+```php
+$basePath = realpath(__DIR__ . '/../../../../');
+require_once $basePath . '/config/config.inc.php';
+```
+
+**Loading Module Classes**: Commands by default won't have module classes loaded. You need to explicitly require the modules before you can use their classes
```
kpodemski:
```suggestion
**Loading Module Classes**: Commands, by default, won't have module classes loaded. You must require the module's main file before using their classes.
```

kpodemski:
Thanks @ioweb-gr @thomasnares 

I've improved it a little and I merge it 👍

PR #1411, opened 2022-07-16
Update jQuery version regarding PrestaShop 1.7.8

PrestaEdit:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | jQuery version was updated inside 1.7.8

### DIFF
--- a/themes/getting-started/theme-organization.md
+++ b/themes/getting-started/theme-organization.md
@@ -127,6 +127,7 @@ add a comment indicating where the code related to products can be found ;)
 
 #### Required libraries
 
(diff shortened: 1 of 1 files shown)


PrestaEdit:
N.B.: This will be needed to be on the 1.7 doc, too. I was on the 8 version while making the change, awaiting the validation of this to backport.

ghost:
Why not just change the version for the doc 8.x?

And for the doc 1.7 make this change.

Also 

`{{< minver v="1.7.8" title="true" >}}` adds a line break.

Screenshot (86)

PrestaEdit:
Thx @okom3pom ; you are right : it will be better as it. Made two another PR's to be clean.

PR #1430, opened 2022-08-02
Use hooks on Symfony page: new catalog-tools button for PS 1.7.6.5, switch to SF 4.4

leemyongpakvn:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Branch?       | 8.x
| Description?  | Since PS 1.7.5.2, catalog-tools-button is toggled instead of a horizontal list <br> Since PS 8.0.0, Symfony has been upgraded to version 4.4
| Fixed ticket? | Fixes #{issue number here} if there is a related issue

PR #1512, opened 2022-09-23
Fix inconsistency link between MarkDown and HugoServer for PR 1430

leemyongpakvn:
./ can not solve inconsistency. We still need 1 level jump up with ../ to make image links work in HugoServer :(

| Questions     | Answers
| Branch?       | 8.x
| Description?  | Fix inconsistency link between MarkDown and HugoServer for PR #1430

### DIFF
--- a/modules/concepts/hooks/use-hooks-on-modern-pages.md
+++ b/modules/concepts/hooks/use-hooks-on-modern-pages.md
@@ -182,7 +182,7 @@ public function hookDisplayDashboardToolbarIcons($hookParams)
 
 In Product Catalog Page you should see the list of Products in debug toolbar in "Dump" section:
 
-![Get products in Dump section](./Catalog_Products_dump.png)
+![Get products in Dump section](../Catalog_Products_dump.png)
 
 #### Using the Symfony components to create an XML export file
 
@@ -284,7 +284,7 @@ And now, the template:
 We have used a key for translation, making our own translations available in back office when using Twig.
 {{% /notice %}}
 
-![Export XML action button](./Catalog_Products_2xml.png)
+![Export XML action button](../Catalog_Products_2xml.png)
 
 And "voila!", the module could be of course improved with so many features, adding filters on export for instance, using the `request` hook parameter and updating the Product repository.
 


kpodemski:
What if we put images in one folder and load them from`/assets/something/something.ext`? I think that would be easier to maintain 🤔

leemyongpakvn:
@kpodemski I use `/assets/modules/concepts_hooks-Catalog_Products_2xml.png` at https://github.com/leemyongpakvn/docs/blob/8.x/modules/concepts/hooks/use-hooks-on-modern-pages.md in line **287**. It works on MarkDown preview. But I quite sure it only works on HugoServer with **/8** prefix as we can see in line **291** below.
Maybe we need tweaks on MarkDown or HugoServer setting.

thomasnares:
Hi [leemyongpakvn](https://github.com/leemyongpakvn), according to : 

[https://gohugo.io/content-management/organization/](https://gohugo.io/content-management/organization/)

Because of the way Hugo resolves urls, if we use an asset on a page that is not an _index.md, we should create a subfolder named like the page, place the assets inside, and move the original page inside subfolder, and change its name to _index.md. 

Then, Hugo will resolve correctly the assets.

In our example : 

```
hooks
├── _index.md
├── list-of-hooks.md
└── use-hooks-on-modern-pages
    ├── Catalog_Products_2xml.png
    ├── Catalog_Products_dump.png
    └── _index.md
```

And since assets belongs (in most cases) to a page, it makes sense to put them in a subfolder.

What do you think ?

leemyongpakvn:
@thomasnares We have dozens of .md pages in both 1.7 and 8 branches so changing them to `_index.md` in a sub folder that way will take a lot of time. Furthermore, if one day, we don't like PrettyURL any more, and switch to UgglyURL like this use-hooks-on-modern-pages`.html`, then we must change page structure again, it will be a pain :(. So I'd prefer using Markdown Render Hooks.

thomasnares:
Ok, makes sense then 👍

thomasnares (approved):

kpodemski (approved):

kpodemski:
Thanks @leemyongpakvn

PR #1558, opened 2022-10-18
Add a doc about Module Self Configurator feature

thomasnares:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Add docs regarding module Self Configurator feature

### DIFF
--- a/modules/configure-with-cli.md
+++ b/modules/configure-with-cli.md
@@ -11,12 +11,12 @@ Self Configurator is a feature that allows module configuration management from 
 
 From a configuration file, the command is able to do some configuration on a specific module, e.g.:
 
-- change values in the Configuration table
+- change values in the `Configuration` table
 - execute SQL statements from a `.sql` file
 - copy files
 - execute `php` scripts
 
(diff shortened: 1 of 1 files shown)


modules/configure-with-cli.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+```yaml
+configuration:
+    delete:
+        - "PAYPAL_ONBOARDING"
+```
+
+{{% notice info %}}
+Please note that you can only `update` or `delete` values: `create` values are only done when installing the module.
```
PrestaEdit: You can add value. The update will create if needed.

thomasnares:
```suggestion
Please note that you can `create` values by specifying them in the `update` section.
```

PrestaEdit (approved):

modules/configure-with-cli.md
thomasnares: thanks, fixed 👍

matks (approved):
Looks great to me thanks @thomasnares

Issue #1436, opened 2022-08-04
Document the use of a new password policy in theme (frontend)


kpodemski:
https://github.com/PrestaShop/PrestaShop/pull/28127/

Base on:
https://github.com/PrestaShop/classic-theme/pull/21

matks:
Assigned to @NeOMakinG 😉

matks:
@kpodemski I think we need to document 'how to use it in a theme' AND 'how it works'

NeOMakinG:
@matks the backend part also needs to be documented and I won't be able to do it!

matks:
@NeOMakinG Yes, I think we need to create a 2nd issue for the backend

PR #1593, opened 2023-01-09
Document new password policy introduced in 8.0

thomasnares:
| Questions     | Answers
| Branch?       | 1.7.x / 8.x
| Description?  | Documentation about the new password policy with zxcvbn introduced in 8.0
| Related PR? | https://github.com/PrestaShop/classic-theme/pull/21 , https://github.com/PrestaShop/PrestaShop/pull/28127 , https://github.com/PrestaShop/docs/issues/1436


thomasnares:
@NeOMakinG you may have suggestions for me about what to document on the frontend part ? 
@PierreRambaud you may also have suggestions for me about backend part 👍

modules/core-updates/8.0.md
```diff
@@ -33,7 +33,7 @@ PrestaShop 8.0 adds supports for PHP 8.0 and PHP 8.1 and requires at least PHP 7
 * Because of the [support for PHP 8.1](https://github.com/PrestaShop/PrestaShop/pull/28402) many methods have declared a return type now.
 * The logic for customers' login and registration [is now](https://github.com/PrestaShop/PrestaShop/pull/27755/) split between two controllers (`RegistrationController`, `AuthController`). This might impact third-party themes and modules as the URL to the registration has changed.
 * Before PrestaShop places an order [there is an additional request now](https://github.com/PrestaShop/PrestaShop/pull/26048/), to see if the number of products in the cart is still valid. This might require third-party payment modules to implement the checks in their solutions accordingly.
-* Due to the new [password policy management features](https://github.com/PrestaShop/PrestaShop/pull/28127), third-party solutions that generate customer data might require to implement changes accordingly.
+* Due to the new [password policy management features](/8/modules/core-updates/new-password-policy), third-party solutions that generate customer data might require to implement changes accordingly.
```
eternoendless: Always use relref for internal links (to ensure they never get broken)

modules/core-updates/new-password-policy.md
```diff
@@ -0,0 +1,67 @@
+---
+title: New password policy based on zxcvbn
+menuTitle: New password policy
+---
+
+# New password policy introduced in 8.0.x
```
eternoendless, line 6: Don't you think it would be more interesting to document this from a system reference standpoint rather from a "how to update your software"? I think it could be overly specific otherwise.

Or perhaps the doc can be split in two: one which explains how the password policy is integrated with customers and employees (for reference), and another one that describes how to update existing code (useful for upgrade only).

NeOMakinG:
> @NeOMakinG you may have suggestions for me about what to document on the frontend part ? @PierreRambaud you may also have suggestions for me about backend part 👍

I guess that's fine enough, we may talk about the fact that `Zxcbn` is loaded async on core.js side because his size is huge but that's it, nothing else to explain

modules/core-updates/new-password-policy.md
thomasnares: Does this look better with the new commit ?

kpodemski (approved):

PR #1608, opened 2023-02-01
Rules for modules technical names

audebms:
Rules for modules technical names

| Questions     | Answers
| Branch?       | 1.7.x / 8.x
| Description?  | Important rules for modules technical names


thomasnares:
Hi @audebms, thanks for your contribution.
This is the same notice as for line 18 (22 now) right ?

audebms:
Almost, I think it is imperative to add that there should be no "_" because PrestaShop modules use this character and this can be confusing.

Maybe my change should be moved to line 22.

thomasnares:
Added a link to explain why it is discouraged in https://github.com/PrestaShop/docs/pull/1621

PR #1700, opened 2023-07-25
About windows and macosx

matks:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | About windows and macosx
| Fixed ticket? |

### DIFF
--- a/basics/installation/system-requirements.md
+++ b/basics/installation/system-requirements.md
@@ -11,7 +11,7 @@ PrestaShop needs the following server configuration in order to run:
 
 | Requirement | Configuration 
 | ---------------- | ------------------
-| **System** | **Unix, Linux**, Windows or MacOS.
+| **System** | **Unix**
 | **Web server** | Apache Web Server 2.4 or any later version. (Although compatible, Apache Web Server 2.2 is not recommended since it [reached its EOL in 2018](https://httpd.apache.org/)).
 | **PHP** | **We recommend PHP 8.1**. See the compatibility chart below for more details. (Although compatible, PHP 7.2.5 to 8.0 are not recommended since they reached or will reach soon their end of support.)
 | **MySQL** | 5.6 minimum, **a recent version is recommended**.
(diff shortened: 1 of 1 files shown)


thomasnares:
I agree that it's best to develop on unix based OS, but do you think that removing windows and macos from the compatibility table risk of discouraging newcomers to the solution ?

matks:
> I agree that it's best to develop on unix based OS, but do you think that removing windows and macos from the compatibility table risk of discouraging newcomers to the solution ?

Possibly, but if we keep them, we kinda lie to users ^^

PrestaShop does not really support Windows or MacOSX. What I mean is: when new releases of PrestaShop (like 8.1.0, 8.1.1, 1.7.8.10 ...) are tested, they are NOT tested with Windows or MacOs X. When PRs are tested, they are NOT tested with Windows or MacOS X. They are always tested with Unix.

(I removed Linux because it is a subset of Unix but we can keep it to make it un-confusing)

Most of the software works with Windows because PHP for Windows behaves very closely to PHP for Linux. So it's more a matter of luck ^^ than a real action from our part.

thomasnares:
From what i just saw, almost all CMS/ecommerce solutions based on php specifies requirements for php, mysql, ... but never for the OS.

Should we go farther and drop the notice about the OS ?

eternoendless:
I suggest we replace the "compatibility" with a note stating that it is recommended to run PrestaShop on linux although it is known to work in other environments.

matks:
> Should we go farther and drop the notice about the OS ?

I like the idea but @Hlavtox is going to say "Eyh! If it's not WRITTEN it does not work Windows then it means it works!" :trollface:

nicosomb:
> When PRs are tested, they are NOT tested with Windows or MacOS X. They are always tested with Unix.

macOS is Unix based 😬

basics/installation/system-requirements.md
```diff
@@ -11,7 +11,7 @@ PrestaShop needs the following server configuration in order to run:
 
 | Requirement | Configuration 
 | ---------------- | ------------------
-| **System** | **Unix, Linux**, Windows or MacOS.
+| **System** | **Unix**
```
kpodemski:
```suggestion
```

kpodemski: Realistically speaking, I don't think we need a system requirement :) PrestaShop needs a web server, and its configuration is the most important. Ok, sometimes you may encounter system-specific issues, but those are not PrestaShop-related problems IMHO.

matks:
> > When PRs are tested, they are NOT tested with Windows or MacOS X. They are always tested with Unix.
> 
> macOS is Unix based 😬

True 👍 nice catch

Hlavtox:
@matks words fly away, written ones remain 😛

matks:
OK guys I think we agree on the idea, but just differ on the details

I don't really care about the final choice of words, I just want people to understand that clarify that Windows is not actively supported by the project. We don't spend time testing with Windows, we don't force ourselves to verify our code changes with Windows.

eternoendless:
I'm in favor of removing the requirement and replacing it by a note explaining that PrestaShop works best in Unix-based systems such as Linux and macOS, and while Windows might work, it is not actively supported.

matks:
I applied @eternoendless suggestion

kpodemski (approved):

Hlavtox (approved):

kpodemski:
thanks @matks :>

PR #1724, opened 2023-09-09
Allow .well-known directory (pki validation / letsencrypt)

BarbUk:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Fix nginx example configuration to allow .well-known directory usage, which is used for pki validation or acme / letsencrypt
| Fixed ticket? | no

### DIFF
--- a/scale/webservers/nginx.md
+++ b/scale/webservers/nginx.md
@@ -107,7 +107,7 @@ server {
 }
 ```
 
-### Allow usage of `.well-known` directory (letsencrypt/pki validation, apple pay)
+### Allow usage of `.well-known` directory (Let's Encrypt / PKI validation, Apple Pay)
 
 The `.well-known` directory is a resource documented in [RFC 8615](https://datatracker.ietf.org/doc/html/rfc8615) and used by [many services](https://en.m.wikipedia.org/wiki/Well-known_URI).
 If you need an external access to the `.well-known` directory, you can update you nginx configuration to use:
--- a/scale/webservers/nginx.md
+++ b/scale/webservers/nginx.md
@@ -106,3 +106,20 @@ server {
     ...
 }
 ```
+
+### Allow usage of `.well-known` directory (letsencrypt/pki validation, apple pay)
+
+The `.well-known` directory is a resource documented in [RFC 8615](https://datatracker.ietf.org/doc/html/rfc8615) and used by [many services](https://en.m.wikipedia.org/wiki/Well-known_URI).
+If you need an external access to the `.well-known` directory, you can update you nginx configuration to use:
+
+```nginx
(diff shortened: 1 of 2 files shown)


BarbUk:
The same modification should be done to the 1.7 branch.
Should I make another PR ?

matks:
> The same modification should be done to the 1.7 branch. Should I make another PR ?

Usually we first merge the PR on 8.x **then we create another PR** and use `git cherry-pick` to backport the commit

cdesiles:
Hi there!
It's interesting to mention nginx can be bound to let's encrypt, but I'm not sure this is relevant as the default example for PrestaShop. 

Don't get me wrong, this improvement could be provided as a secondary integration example, but you never know the will of the user integrating nginx + prestashop. Use cases are very wide, it could be placed behind another reverse-proxy taking care of the HTTPS, or any other http component... Better not document (and maintain) what is not closely relevant to PrestaShop IMO.

matks (changes requested):
I'm not a NGINX expert but I trust @jokesterfr on this topic 👍 

The developer documentation of PrestaShop must act as a standard guide for prestashop newcomers so we can only add in it generic and standard information, we can't push side-cases

BarbUk:
Letsencrypt is one of the [many services](https://en.m.wikipedia.org/wiki/Well-known_URI) that use a well-known dir. 

Usage of this resource is documented in [RFC 8615](https://datatracker.ietf.org/doc/html/rfc8615).

You document a configuration that explicitly deny usage of this directory. It should not be blocked by default IMHO.

thomasnares:
Hi @BarbUk, i suggest moving your example to: https://devdocs.prestashop-project.org/8/scale/webservers/nginx/

- the page "installation" is more about local environment than internet exposed PS instance; 
- the page "Optimize your nginx configuration" is more about going to production, so it makes sense to mention .well-known here. 

You could create a heading and show your code snippet 👍

matks:
Great suggestion from @thomasnares

BarbUk:
Thanks for the suggestion @thomasnares.
I've updated the PR.

scale/webservers/nginx.md
```diff
@@ -106,3 +106,20 @@ server {
     ...
 }
 ```
+
+### Allow usage of `.well-known` directory (letsencrypt/pki validation, apple pay)
```
thomasnares:
```suggestion
### Allow usage of `.well-known` directory (Let's Encrypt / PKI validation, Apple Pay)
```

thomasnares (approved):
Thanks @BarbUk, i just added some majs on the brands/services

matks (approved):

kpodemski:
thank you @BarbUk !

PR #1741, opened 2023-10-30
Script to extract hooks

thomasnares:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Cleaned the script to generate hook list, but there are still some bugs on regex. 

**This script allows to extract hooks for documentation for devdocs, from a release archive, or from a directory**

Usage is documented within the script

But it must be finished and fixed: 

## Issues with this script: 

- [ ] Fix some regexes to match hooks in codebase (you must scan all files, and identify where hooks are called)
- [ ] Fix the name cleaning array (linked with regex matching)
- [ ] Detect if hooks uses `check_exceptions`, `chain` and `array_return`

**I'm still not sure we can match all hooks with regexes against the codebase.** 

## Solution

Another approach would be to refactor the `HookFinder` with the famous php-parser library which produces AST : https://github.com/nikic/php-parser (the one used by PHPStan to detect bugs). 

As the hook list is in perpetual evolution and changes at every release, this should be a priority to fix this script and use it regularely

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ This referential can be generated with a script, crawling all:
 - all native modules
 - all native themes
 
+An example of such a script can be found in this repository, in: `hookExctractionScript.php`.
+
+Usage of this script: `php hookExtraction.php --source=prestashop-release.zip --versionCode=8.1.2 --outputFolder=modules/concepts/hooks/list-of-hooks`
+
 In the documentation, a hook is described in yaml format: 
 
 ```yaml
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -34,6 +34,10 @@ This referential can be generated with a script, crawling all:
 - all native modules
 - all native themes
 
+An example of such a script can be found in this repository, in: `hookExctractionScript.php`.
```
thomasnares:
```suggestion
An example of such a script can be found in this repository, in: `hookExtractionScript.php`.
```

kpodemski:
Closed in favor of the final version of the script prepared by @tleon 

From one Thomas to another! :D thank you to you both!

PR #1757, opened 2023-11-27
Specific vertion of PS where it is available

msaustral:
| Questions     | Answers
| Branch?       | 1.7.x / 8.x

### DIFF
--- a/development/components/form/types-reference/number-min-max-filter.md
+++ b/development/components/form/types-reference/number-min-max-filter.md
@@ -1,3 +1,5 @@
+ONLY AVAILABLE FROM PS 1.7.7.X ->
+
 ---
 title: NumberMinMaxFilterType
 ---


development/components/form/types-reference/number-min-max-filter.md
```diff
@@ -1,3 +1,5 @@
+ONLY AVAILABLE FROM PS 1.7.7.X ->
```
thomasnares, line 1: We should not indicate this information here, but in the first heading of the page, with the specific markdown

```
{{< minver v="1.7.7" >}}
```

thomasnares (commented):
Hi @msaustral, thanks for your contribution, i left you a comment 👍 

Can you put it on the H1 line of the page (line 7)

kpodemski (changes requested):
Hello @msaustral 

Please follow @thomasnares suggestions.

kpodemski:
We have this information in ps8 docs, I think we can close this one.

PR #1764, opened 2023-12-14
fix(translation domains): Update case of module name in translations

Kaikina:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Modules name in translation domains should not use camel case.
| Fixed ticket? | NA


PululuK (approved):

PululuK:
Thanks @Kaikina

PR #1783, opened 2024-02-25
Update new-system.md

Vots1000:
Changing the order of arguments in the symfony example.

| Questions     | Answers
| Branch?       | 1.7.x / 8.x
| Description?  | Changing the order of arguments for trans() function in the symfony system example.
| Fixed ticket? |


kpodemski (approved):
Nice catch! We indeed have a different parameters order here and inside `ModuleFrontController` 😅

kpodemski:
thanks @Vots1000 :)

PR #1806, opened 2024-05-07
change php version for 9

jf-viguier:
| Questions     | Answers
| Branch?       | 9
| Description?  | php version


kpodemski:
@PrestaEdit was faster this time Jeff :)

jf-viguier:
@kpodemski ideed ! 
I've set the recommanded version to php 8.3, @PrestaEdit 8.1, were is the truth ?

PrestaEdit:
On my own, I will suggest the 8.1 as the support of the two others versions are added after.
Also, the latest 1.7.8.x are made to support the 8.1 at maximum. It will be a good interesection.

kpodemski:
@jf-viguier @PrestaEdit 

I don't like such recommendations, it all depends on various factors and common sense would tell that the best is the recent version, but my personal opinion is similar to yours, Jonathan. I think it would be wiser to expect ecosystem to support PHP 8.1, PHP 8.2 and 8.3 may require some extra effort.

PR #1823, opened 2024-06-12
Detail the new login migration impacts

jolelievre:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Detail the new login migration impacts
| Fixed ticket? | ~

### DIFF
--- a/modules/core-updates/9.0.md
+++ b/modules/core-updates/9.0.md
@@ -85,6 +85,68 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### BO login and authorization migrated to Symfony
+
+The back office login page has been migrated to Symfony. Along with this change, the authorization system in the back office is now also based on Symfony, which implies several things:
+- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side, and very few data are kept on the browser side. For retro-compatibility, we still populate the legacy cookie so that you can **read** it. While you can use `Context::$cookie` for your custom data, making changes to the data previously used for PrestaShop authorization will no longer have the desired effect and may even result in instability, so it's not recommended
+- the `PrestaShopBundle\Security\Admin\EmployeeProvider` and `PrestaShopBundle\Service\DataProvider\UserProvider` now return a `PrestaShopBundle\Entity\Employee` instance, their responsibility is to return a `Symfony\Component\Security\Core\User\UserInterface` anyway (and they still do) but in case you depended on the child class know that `PrestaShopBundle\Security\Admin\Employee` no longer exist
+- if you need to get the logged-in user, you can use the Symfony `Symfony\Bundle\SecurityBundle\Security` service, but we recommend you use the `EmployeeContext` internal PrestaShop service (see explanation about [new Contexts]({{< relref "9/modules/core-updates/9.0.md#symfony-context-refacto" >}}))
+
+**Storing custom data in Session**
+
+If you need to persist some custom Employee data, we don't recommend using the `Context::$cookie` anymore, instead you can use the `Session` from Symfony and update its attributes:
+
+```php
+// In Symfony controllers, you can get the session from the Request parameter injected by Symfony
+$session = $request->getSession();
+
+// In other services you will need to get the current request via the RequestStack service
+$request = $requestStack->getCurrentRequest();
+if ($request) {
+  $session = $request->getSession();
+}
+
+// You can then get/set a custom attribute, but you should always check that the session object is indeed available. Some requests do not rely on the session, and it may never be created
+if ($session) {
+  // Be careful to use a custom and UNIQUE attribute name
+  $myData = $session->getAttribute('my_custom_data');
+  // The session object will be automatically serialized at the end of the request, and automatically unserialized at the beginning of the next requests
+  $session->setAttribute('my_custom_data', 'updated value');
+}
+```
+
+{{% notice note %}}
+The Symfony session has other interesting features (like flash bags for unique usage data). You can read more about its usage in the <a href="https://symfony.com/doc/6.4/session.html" target="_blank">Session documentation</a>.
+{{% /notice %}}
+
+
+**Removed hooks:**
+- `actionAdminLoginControllerBefore`
+- `actionAdminLoginControllerLoginBefore`
+- `actionAdminLoginControllerLoginAfter`
+- `actionAdminLoginControllerForgotBefore`
+- `actionAdminLoginControllerForgotAfter`
+- `actionAdminLoginControllerResetBefore`
+- `actionAdminLoginControllerResetAfter`
+
+**Hooks kept for backward compatibility:**
+- `actionAdminLoginControllerSetMedia` so you can add some custom assets on the login page
+- `displayAdminLogin` so you can add custom HTML on the login page
+
+**New hooks**
+- `actionBackOfficeLoginForm` to modify the login form (the form builder is passed via the `form_builder` hook parameter)
+- `actionEmployeeRequestPasswordResetForm` to modify the request password form (the form builder is passed via the `form_builder` hook parameter)
+
+The login logic, however, is not handled by PrestaShop's internal code so there is no handler that interprets the submitted data (and that's why there are so few new hooks). Instead we rely on <a href="https://symfony.com/doc/current/security.html#form-login" target="_blank">Symfony form login authentication system</a>, so it's Symfony's internal system that interprets the data submitted, validates it and authenticate the employee to the back office.
+
+If you need to integrate your code with the authentication process you should now rely on the <a href="https://symfony.com/doc/current/security.html#authentication-events" target="_blank">Authentication events</a> described in their documentation, you can also inspire yourself from our <a href="https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/EventListener/Admin/EmployeeSessionSubscriber.php" target="_blank">own internal subscriber</a>.
+
+**Related PRs:**
+- [Update employee interfaces](https://github.com/PrestaShop/PrestaShop/pull/34818)
+- [PS_SSL_ENABLED_EVERYWHERE configuration removed, PS_TRUSTED_PROXIES env variable introduced](https://github.com/PrestaShop/PrestaShop/pull/35761)
+- [AdminController::isAnonymousAllowed is now public](https://github.com/PrestaShop/PrestaShop/pull/35810)
+- [Login page and authentication migrated](https://github.com/PrestaShop/PrestaShop/pull/35983)
+
 ## Migrated pages
 
 The migration of pages imply several breaking changes, we won't detail all of them for each page but here is a summary:
@@ -135,15 +197,6 @@ This changes the data passed to the smarty templates:
 - [Removed modal_module_list, modals views in new layout](https://github.com/PrestaShop/PrestaShop/pull/33864)
 - [Clean Smarty variables that were only relevant for the layout internal use](https://github.com/PrestaShop/PrestaShop/pull/34560)
 - [Legacy pages are rendered by Symfony](https://github.com/PrestaShop/PrestaShop/pull/34783)
-
-### BO login and authorization migrated to Symfony
-
-@todo detail the impact of this migration (legacy Cookie no longer usable, CQRS token always used in urls, session stored on server side, ...)
-
-- [Update employee interfaces](https://github.com/PrestaShop/PrestaShop/pull/34818)
-- [PS_SSL_ENABLED_EVERYWHERE configuration removed, PS_TRUSTED_PROXIES env variable introduced](https://github.com/PrestaShop/PrestaShop/pull/35761)
-- [AdminController::isAnonymousAllowed is now public](https://github.com/PrestaShop/PrestaShop/pull/35810)
-- [Login page and authentication migrated](https://github.com/PrestaShop/PrestaShop/pull/35983)
 
 ### Symfony context refacto
 


modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### BO login and authorization migrated to Symfony
+
+The login page to the back-office has been migrated to Symfony, along ith it the system of authorization in the BackOffice is now also based on Symfony, it implies several things:
+- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side and very few data are kept on the browser side, for retro-compatibility we still populate the legacy cookie so that you can **read** it but modifying/writing it will likely no longer have the effect you would expect (if any)
```
Hlavtox: @jolelievre Jo, will this $context->cookie still be accessible or not? Because some things depend on it, For example - https://github.com/PrestaShop/ps_googleanalytics/blob/618ade8f9a424ad31968f6367ddf3e0be8c5f879/ps_googleanalytics.php#L131

If not, could you note some other way to store temporary data between sessions? ($session?)

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
...
+
+**Hooks kept for backward compatibility:**
+- `actionAdminLoginControllerSetMedia` so you can add some custom assets on the login page
+- `displayAdminLogin` so you can add custom HTML on the login page
+
+**New hooks**
+- `actionBackOfficeLoginForm` to modify the login form (the form builder is passed via the `form_builder` hook parameter)
+- `actionEmployeeRequestPasswordResetForm` to modify the request password form (the form builder is passed via the `form_builder` hook parameter)
+
+The logic of the login, however, is not handled by PrestaShop internal code so there is no handler that interprets the submitted data (and that's hy there is so few new hooks), instead we rely on <a href="https://symfony.com/doc/current/security.html#form-login" target="_blank">Symfony form login authentication system</a>, so it's Symfony internal system that interprets the data submitted, validates it and authenticate the employee to the back-office.
```
PrestaEdit:
```suggestion
The logic of the login, however, is not handled by PrestaShop internal code so there is no handler that interprets the submitted data (and that's why there is so few new hooks), instead we rely on <a href="https://symfony.com/doc/current/security.html#form-login" target="_blank">Symfony form login authentication system</a>, so it's Symfony internal system that interprets the data submitted, validates it and authenticate the employee to the back-office.
```

modules/core-updates/9.0.md
jolelievre: Hmm actually you can still write data in the cookie, it's just that the core no longer depends on it So for example you can't perform a BO login by manipulating the cookie anymore
But persisted data is still present in the cookie so it should be compatible for the modules, but you're right I'm gonna detail this a bit more

And regarding new possibilities to persist data, you can save attributes in the Symfony token That's what we did to handle:
- shop context
- employee session
- IP address checking

So it's worth detailing this as well I guess.

matthieu-rolland (approved):

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### BO login and authorization migrated to Symfony
+
+The login page to the back-office has been migrated to Symfony, along ith it the system of authorization in the BackOffice is now also based on Symfony, it implies several things:
```
```suggestion
The back office login page has been migrated to Symfony. Along with this change, the authorization system in the back office is now also based on Symfony, which implies several things:
```

modules/core-updates/9.0.md
```suggestion
- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side, and very few data are kept on the browser side, for retro-compatibility we still populate the legacy cookie so that you can **read** it, but modifying/writing it will likely no longer have the effect you would expect (if any)
```

as @Hlavtox mentioned, it requires some clarification because it is probably not true for most of the scenarios cookies have been used by community developers

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side and very few data are kept on the browser side, for retro-compatibility we still populate the legacy cookie so that you can **read** it but modifying/writing it will likely no longer have the effect you would expect (if any)
+- the `PrestaShopBundle\Security\Admin\EmployeeProvider` and `PrestaShopBundle\Service\DataProvider\UserProvider` now return a `PrestaShopBundle\Entity\Employee` instance, their responsibility as to return a `Symfony\Component\Security\Core\User\UserInterface` anyway (and they still do) but in case you depended on the child class know that `PrestaShopBundle\Security\Admin\Employee` no longer exist
```
```suggestion
- the `PrestaShopBundle\Security\Admin\EmployeeProvider` and `PrestaShopBundle\Service\DataProvider\UserProvider` now return a `PrestaShopBundle\Entity\Employee` instance, their responsibility is to return a `Symfony\Component\Security\Core\User\UserInterface` anyway (and they still do) but in case you depended on the child class know that `PrestaShopBundle\Security\Admin\Employee` no longer exist
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- if you need to get the logged-in user you can use the Symfony `Symfony\Bundle\SecurityBundle\Security` service, but we recommend you'd rather use the `EmployeeContext` internal PrestaShop service instead (see explanation about [new Contexts]({{< relref "9/modules/core-updates/9.0.md#symfony-context-refacto" >}}))
```
```suggestion
- if you need to get the logged-in user, you can use the Symfony `Symfony\Bundle\SecurityBundle\Security` service, but we recommend you use the `EmployeeContext` the internal PrestaShop service (see explanation about [new Contexts]({{< relref "9/modules/core-updates/9.0.md#symfony-context-refacto" >}}))
```

modules/core-updates/9.0.md
```suggestion
The login logic is not handled by PrestaShop's internal code, so there is no handler that interprets the submitted data (and that's why there are so few new hooks). Instead, we rely on <a href="https://symfony.com/doc/current/security.html#form-login" target="_blank">Symfony form login authentication system</a>, so it's Symfony's internal system that interprets the data submitted, validates it and authenticate the employee to the back office.
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- `actionBackOfficeLoginForm` to modify the login form (the form builder is passed via the `form_builder` hook parameter)
+- `actionEmployeeRequestPasswordResetForm` to modify the request password form (the form builder is passed via the `form_builder` hook parameter)
+
+The logic of the login, however, is not handled by PrestaShop internal code so there is no handler that interprets the submitted data (and that's hy there is so few new hooks), instead we rely on <a href="https://symfony.com/doc/current/security.html#form-login" target="_blank">Symfony form login authentication system</a>, so it's Symfony internal system that interprets the data submitted, validates it and authenticate the employee to the back-office.
+
+If you need to interface your code with the authentication process you should now rely on the <a href="https://symfony.com/doc/current/security.html#authentication-events" target="_blank">Authentication events</a> described in their documentation, you can also inspire yourself from our <a href="https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/EventListener/Admin/EmployeeSessionSubscriber.php" target="_blank">own internal subscriber</a>.
```
```suggestion
If you need to interfere with the authentication process, you should now rely on the <a href="https://symfony.com/doc/current/security.html#authentication-events" target="_blank">Authentication events</a> described in their documentation, you can also inspire yourself from our <a href="https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/EventListener/Admin/EmployeeSessionSubscriber.php" target="_blank">own internal subscriber</a>.
```

kpodemski (commented):
A few comments.

Any particular reason not to use Markdown URLs? Is it because of the `target="_blank"`?

jolelievre:
> A few comments.
> 
> Any particular reason not to use Markdown URLs? Is it because of the `target="_blank"`?

Yep, I couldn't find a way to have target blank with the markdown syntax, and for external documentation I thought it was better to open them in another tab.

modules/core-updates/9.0.md
jolelievre: Yeah I modified this, let me know after I push my changes it it's clearer

modules/core-updates/9.0.md
jolelievre: I prefered the word `integrate`, interfere makes it feel like you wanna prevent the initial behaviour more than add something to it

jolelievre:
@Hlavtox @kpodemski do you think it's clearer now?

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,40 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### BO login and authorization migrated to Symfony
+
+The back office login page has been migrated to Symfony. Along with this change, the authorization system in the back office is now also based on Symfony, which implies several things:
+- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side, and very few data are kept on the browser side, for retro-compatibility we still populate the legacy cookie so that you can **read** it Although you can modify it for your custom data but modifying the data previously used for PrestaShop authorization will no longer have effect (or unstable ones) so it's not recommended
```
```suggestion
- we no longer depend on the legacy `Context::$cookie`, the session is kept on the server side, and very few data are kept on the browser side. For retro-compatibility, we still populate the legacy cookie so that you can **read** it. While you can work with `Context::$cookie` and your custom data, making changes to the data used for PrestaShop authorization may no longer have the desired effect or may result in instability, so it's not recommended.
```

A little rephrase.

Hlavtox:
@jolelievre Yes, all clear now! :-)

-------------

By the way, is the recommended way to store some temporary data the symfony session?
If yes, we can add a new section in the docs and link it. Something like:

In the past, if you needed to store lightweight temporary data for the session, you could use the Prestashop cookie for this purpose. However, with the recent migration of login system to symfony, this is slowly getting obsolete. It was never a good solution in a first place, because there are risks with overflowing the cookies and other issues.

So, if you want save something temporarily in between of requests:
- Get `$requestStack` either by DI or by using `PrestaShop\PrestaShop\Adapter\SymfonyContainer::getInstance()->get('request_stack')` in legacy environment.
- Store data by using `$requestStack->getSession()->set('key', $value);`
- Get data by using `$requestStack->getSession()->get('key');`

jolelievre:
@Hlavtox the temporary persistence you are defining is not related to the login/authorization migration Using session between requests was always possible and still is, but it's only stored for ONE request, I prefer not to confuse this part of the doc with something unrelated.

However, I'm gonna add a paragraph about persisting data with the Token, which is an alternative to using the legacy Cookie and is more in tune with this feature

jolelievre:
@Hlavtox Actually you're right, session is persisted between multiple requests I was mixing things with flash bags that only last one request I'll follow your suggestion as it's probably safer to encourage people to use the session rather than the token attributes as we do internally In our case I think it's fine because we store data relevant for the authorization process, but it shouldn't be used for more generic purposes

jolelievre:
@Hlavtox @kpodemski this should be good, with extra explanation on how to use the session, this page is getting a bit big but I want to finish each section first and once it's done the page will likely be reworked Some of these sections will probably be moved in a dedicated page

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,68 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- the `PrestaShopBundle\Security\Admin\EmployeeProvider` and `PrestaShopBundle\Service\DataProvider\UserProvider` now return a `PrestaShopBundle\Entity\Employee` instance, their responsibility is to return a `Symfony\Component\Security\Core\User\UserInterface` anyway (and they still do) but in case you depended on the child class know that `PrestaShopBundle\Security\Admin\Employee` no longer exist
+- if you need to get the logged-in user, you can use the Symfony `Symfony\Bundle\SecurityBundle\Security` service, but we recommend you use the `EmployeeContext` internal PrestaShop service (see explanation about [new Contexts]({{< relref "9/modules/core-updates/9.0.md#symfony-context-refacto" >}}))
+
+**Storing custom data in Session**
+
+If you need to persist some custom Employee data we don't recommend using the `Context::$cookie` anymore, instead you can use the `Session` from Symfony and update its attributes:
```
```suggestion
If you need to persist some custom Employee data, we don't recommend using the `Context::$cookie` anymore, instead you can use the `Session` from Symfony and update its attributes:
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,68 @@ Some dependencies are still present but were upgraded which comes with their own
...
+// In Symfony controllers, you can get the session from the Request parameter injected by Symfony
+$session = $request->getSession();
+
+// In other services you will need to get the current request via the RequestStack service
+$request = $requestStack->getCurrentRequest();
+if ($request) {
+  $session = $request->getSession();
+}
+
+// You can then get/set a custom attribute, but you should always check that the session object is indeed available Some request do not rely on session and the session may never be created
```
```suggestion
// You can then get/set a custom attribute, but you should always check that the session object is indeed available. Some requests do not rely on the session, and the session may never be created.
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,68 @@ Some dependencies are still present but were upgraded which comes with their own
...
+if ($session) {
+  // Be careful to use a custom and UNIQUE attribute name
+  $myData = $session->getAttribute('my_custom_data');
+  // The session object will be automatically serialized at the end of the request, and automatically unserialized at the beginning of next requests
```
```suggestion
  // The session object will be automatically serialized at the end of the request, and automatically unserialized at the beginning of the next requests
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,68 @@ Some dependencies are still present but were upgraded which comes with their own
...
+  $session->setAttribute('my_custom_data', 'updated value');
+}
+```
+
+{{% notice note %}}
+The Symfony session has other interesting features (like flash bags for unique usage data), feel free to read more about is usage in the <a href="https://symfony.com/doc/6.4/session.html" target="_blank">Session documentation</a>.
```
```suggestion
The Symfony session has other interesting features (like flash bags for unique usage data). You can read more about its usage in the <a href="https://symfony.com/doc/6.4/session.html" target="_blank">Session documentation</a>.
```

kpodemski:
@jolelievre nice 👍🏻

Hlavtox (approved):
Good job, you're the boss 🎉

jolelievre:
Last changes just pushed, @kpodemski just waiting for your approval and I think we can merge this one
I will then rebase the next one https://github.com/PrestaShop/docs/pull/1827

kpodemski (approved):
You have my approval @jolelievre :D

PR #1827, opened 2024-06-18
Explain new Symfony kernel split

jolelievre:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Explain new Symfony kernel split
| Fixed ticket? | ~

### DIFF
--- a/modules/core-updates/9.0.md
+++ b/modules/core-updates/9.0.md
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### Kernel modifications
+
+Until PrestaShop 8.0, we relied on one Kernel used for the back office. In Symfony 9.0, we expanded the usage of the Symfony framework, especially for two new features:
+- the new Admin API based on API Platform and OAuth authentication
+- an experimental feature that gives access to a Symfony container in the front office (in opposition to the custom-built container available today). This one is not very much used yet, but it's a first stone for the future front office migration
+
+Those two new environments have mechanisms and configuration different from those in the back office. To separate these configurations cleanly, they each have their dedicated kernel class and their own configuration. This allows to cleanly define different routing, security configurations, dedicated services, listeners, ...
+Each kernel has a unique `applicationId` that allows to dynamize its configuration and cache folder (since the services are not the same, each kernel needs its own cache folder). The `appId` parameter has also been added globally to the `bin/console` tool. They all share the common `app/config/config.yml` configuration file, but each one extends it in its own config folder.
+
+| Environment                 | Kernel class     | Config folder          | Cache folder                          | Endpoint               | Application ID |
+|-----------------------------|------------------|------------------------|---------------------------------------|------------------------|----------------|
+| Back office                 | `AdminKernel`    | `app/config/admin`     | `var/cache/{dev,prod,test}/admin`     | `/admin-dev/index.php` | `admin`        |
+| Admin API                   | `AdminAPIKernel` | `app/config/admin-api` | `var/cache/{dev,prod,test}/admin-api` | `/admin-api/index.php` | `admin-api`    |
+| Front office (experimental) | `FrontKernel`    | `app/config/front`     | `var/cache/{dev,prod,test}/front`     | `/index.php`           | `front`        |
+
+{{% notice note %}}
+**Console usage**
+
+Clear the cache of the Admin API for its `prod` environment
+
+`php bin/console cache:clear --env=prod --app-id=admin-api`
+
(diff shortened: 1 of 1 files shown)


Hlavtox:
@jolelievre Jo could you mention to what category the webservice container goes to?

https://github.com/PrestaShop/PrestaShop/blob/b7e6e6e7d87528e87d0eea4fac46b89123991ca1/webservice/dispatcher.php#L39

jolelievre:
> @jolelievre Jo could you mention to what category the webservice container goes to?
> 
> https://github.com/PrestaShop/PrestaShop/blob/b7e6e6e7d87528e87d0eea4fac46b89123991ca1/webservice/dispatcher.php#L39

As far as I know none The webservice endpoint doesn't rely on Symfony, no kernel is instanciated it's purely legacy and cannot use any Symfony feature, or at least I don't find any place in the code that would indicate this 🤔

jolelievre:
@Hlavtox ah actually there is a container instanciated but it's a legacy container, the same kind as the one used currently on the front office. It's custom built and it's not handled by the Symfony framework so it's not relevant in this case IMHO.

This documentation should already explain this https://devdocs.prestashop-project.org/8/modules/concepts/services/#services-in-legacy-environment

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
 - [Services translator, security.csrf.token_manager, session, session.flash_bag no longer public](https://github.com/PrestaShop/PrestaShop/pull/35069)
 - [Replace AdminSecurity annotation with attributes](https://github.com/PrestaShop/PrestaShop/pull/35212)
 
+### Kernel modifications
+
+Until PrestaShop 8.0 we relied on one Kernel that was used for the back office, in Symfony 9.0 we expanded the usage of the Symfony framework especially for two new features:
```
```suggestion
Until PrestaShop 8.0, we relied on one Kernel used for the back office. In Symfony 9.0, we expanded the usage of the Symfony framework, especially for two new features:
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- the new Admin API based on ApiPlatform and OAuth authentication
```
```suggestion
- the new Admin API based on API Platform and OAuth authentication
```

API Platform is an official name :D

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
...
+- an experimental feature that gives access to a Symfony container on the frontend (in opposition to the custom built built container available today), this one is not very much used but it's a first stone for the future front migration
```
```suggestion
- an experimental feature that gives access to a Symfony container in the front office (in opposition to the custom-built container available today). This one is not very much used, but it's a first stone for the future front office migration
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
...
+
+Those two new environments have different mechanisms and configuration from the back office, to separate these configurations cleanly they each have their dedicated kernel class and their own configuration. This allows to cleanly define different routing, security configurations, dedicated services, listeners, ...
```
```suggestion
Those two new environments have mechanisms and configurations different from those in the back office. To separate these configurations cleanly, they each have their dedicated kernel class and their own configuration. This allows to cleanly define different routing, security configurations, dedicated services, listeners, ...
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
...
+Each kernel has a unique `applicationId` that allows to dynamize its configuration, its cache folder (since the services are not the same each kernel needs its own cache folder). The `appId` parameter has also been added globally to the `bin/console` tool. They all share the common `app/config/config.yml` configuration file but each one extends it in its own config folder.
```
```suggestion
Each kernel has a unique `applicationId` that allows it to dynamize its configuration and cache folder (since the services are not the same, each kernel needs its own cache folder). The `appId` parameter has also been added globally to the `bin/console` tool. They all share the common `app/config/config.yml` configuration file, but each one extends it in its own config folder.
```

modules/core-updates/9.0.md
```diff
@@ -85,6 +85,38 @@ Some dependencies are still present but were upgraded which comes with their own
...
+| Front office (experimental) | `FrontKernel`    | `app/config/front`     | `var/cache/{dev,prod,test}/front`     | `/index.php`           | `front`        |
+
+{{% notice note %}}
+**Console usage**
+
+Clear the cache of the Admin API for its `prod` environment
+
+`php bin/console cache:clear --env=prod --app-id=admin-api`
+
+Display event listeners for the Admin in `dev` environment (default value of `app-id` is `admin` for retro compatibility)
```
```suggestion
Display event listeners for the back office `dev` environment (default value of `app-id` is `admin` for retro compatibility)
```

kpodemski:
I added some suggestions @jolelievre

modules/core-updates/9.0.md
jolelievre: Yep you're right!

jolelievre:
Thanks for the review @kpodemski it should be good

kpodemski (approved):

kpodemski:
thanks @jolelievre

PR #1828, opened 2024-06-19
Document context refacto and modifications regarding symfony controllers

jolelievre:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Document context refacto and modifications regarding symfony controllers
| Fixed ticket? | ~

### DIFF
--- a/modules/core-updates/9.0.md
+++ b/modules/core-updates/9.0.md
@@ -78,7 +78,92 @@ Some dependencies are still present but were upgraded which comes with their own
 
 ### Symfony controllers
 
-@todo: explain changes about DI (controllers as services, injected container no longer the same, ...), new PrestaShopAdminController and deprecated FrameworkAdminBundleController
+In Symfony 6.4, the controllers must now be defined as services. The impact on the existing controllers is mainly around the concept of Dependency Injection and how services are injected or accessed in the controllers.
+The container passed to the controllers is no longer the "global container" that contains all the existing services in the application. Instead, it injects a dedicated container optimized for the controller based on the services that are injected into it.
+
+The issue is that most of PrestaShop's Symfony controllers do not rely on injection but instead use the `$this->get('service_name')` method to access any service. This is no longer possible in modern controllers because of the mentioned optimization related to container build (and the `get` method was even removed).
+To avoid a very big breaking change, we modified the `PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController`, so that it can fetch services from the controller's container **and** from the global one, but this is a little hack that goes against Symfony recommendations, as such this base class is already **deprecated and will be removed in PrestaShop 10.0**.
+
+This gives developers some time to refacto their controllers. They should now rely on the new `PrestaShopBundle\Controller\Admin\PrestaShopAdminController`, which doesn't implement the mentioned hack in order to force a proper implementation of the controllers.
+You can read more details about <a href="https://symfony.com/doc/6.4/controller/service.html" target="_blank">Controllers as services</a>, <a href="https://symfony.com/doc/6.4/service_container/service_subscribers_locators.html" target="_blank">Service Subscribers & Locators</a> or <a href="https://symfony.com/doc/6.4/service_container.html">Service container</a> in general.
+
+As an example, you have three main possibilities to inject a service into your controller:
+
+```php
+<?php
+
+use PrestaShopBundle\Controller\Admin\PrestaShopAdminController;
+
+class MyController extends PrestaShopAdminController
+{
+    /**
+     * You can inject your service systematically in the constructor, useful if the service is used by several actions of your controller.
+     * 
+     * @param MyCustomService $myCustomService
+     */
+    public function __construct(
+      private readonly MyCustomService $myCustomService,
+    ) {
+    }
+    
+    /**
+     * You can inject a service only on one method, useful if the service is used only in one action. It is more optimized than injecting it all the time and never using it.
+     * 
+     * @param MySpecificService $specificService
+     * @return JsonResponse
+     */
+    public function indexAction(MySpecificService $specificService): JsonResponse
+    {
+        $generalData = $this->myCustomerService->getData();
+        $specificData = $specifiService->getSpecificData();
+
+        // You can use the get method to fetch a service that was added in the getSubscribedServices method
+        $toolingService = $this->container->get(ToolingService::class);
+        
+        return new JsonResponse($toolingService->format($generalData + $specificData));
+    }
+    
+    /**
+     * You can define a list of registered services, they will be accessible via the $this->container->get method, this
+     * way of injecting service is interesting if implemented on a base controller class for generic services that
+     * may be used by many controllers (like the router, the translator, ...) so you don't need to inject them manually
+     * in each controller class.
+     * 
+     * @return array
+     */
+    public static function getSubscribedServices(): array
+    {
+        return parent::getSubscribedServices() + [
+            ToolingService::class => ToolingService::class,
+        ];
+    }
+}
+```
+
+**Define your controller as a service**
+
+You can already anticipate and define your controllers as services, it is also compatible with previous versions of Symfony and PrestaShop:
+
+```yaml
+services:
+  MyNamespace\MyController:
+    # Using autowire will inject services in the constructors automatically
+    autowire: true
+    # Using autoconfigure with controllers that extend AbstractController will handle method parameter injection and subscribed services
+    autoconfigure: true
+    # If your controller class doesn't extend AbstractController you should add this tag manually so you benefit from method parameters injection
+    tags: ['controller.service_arguments']
+```
+
+{{% notice note %}}
+**PrestaShopAdminController::getSubscribedServices**
+
+The new base controller class uses the `getSubscribedServices` method to give easy access to some commonly used PrestaShop services, like the `ConfigurationInterface`, `HookDispatcherInterface`, `TranslatorInterface`, ...
+You can see the <a href="https://github.com/PrestaShop/PrestaShop/blob/9.0.0-alpha.1/src/PrestaShopBundle/Controller/Admin/PrestaShopAdminController.php#L59" target="_blank">full list of subscribed services</a> in the class, this list will probably grow piece by piece but only for common generic services.
+<br/><br/>
+Thanks to this, you can use `$this->container->get(ConfigurationInterface::class)->get('my_config')` to access the configuration service, or alternatively use the helper method `$this->getConfiguration()->get('my_config')`.
+The `PrestaShopAdminController` class comes with other helpful helper methods that are commonly used in controllers.
+{{% /notice %}}
 
 - [Removed method renderForm which is now used by the Symfony framework](https://github.com/PrestaShop/PrestaShop/pull/31994)
 - [Maintain backward compatibility FrameworkBundleAdminController, but deprecated](https://github.com/PrestaShop/PrestaShop/pull/32450)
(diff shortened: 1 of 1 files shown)


modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
 
 ### Symfony controllers
 
-@todo: explain changes about DI (controllers as services, injected container no longer the same, ...), new PrestaShopAdminController and deprecated FrameworkAdminBundleController
+In Symfony 6.4 the controllers must now be defined as services, the impact for the existing controllers is mainly around the concept od Dependency Injection and how services are injected or accessed in the controllers.
```
```suggestion
In Symfony 6.4, the controllers must now be defined as services. The impact on the existing controllers is mainly around the concept of Dependency Injection and how services are injected or accessed in the controllers.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+The container that is passed in the controllers is no longer the "global container" that contains all the existing services in the application, instead it injects a dedicated container optimized for the controller based on the services that are injected into it.
```
```suggestion
The container passed to the controllers is no longer the "global container" that contains all the existing services in the application. Instead, it injects a dedicated container optimized for the controller based on the services that are injected into it.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+
+The problem is that most of our Symfony controllers don't rely on injection but use the `$this->get('service_name')` method to get any service, this is no longer possible in modern controllers because of the mentioned optimisation related to container build.
```
```suggestion
The issue is that most of PrestaShop's Symfony controllers do not rely on injection but instead use the `$this->get('service_name')` method to access any service. This is no longer possible in modern controllers because of the mentioned optimization related to container build.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+To avoid a very big breaking change we modified the `PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController` so that it is able to fetch services from the controller's container AND from the global one, but this is a little hack that goes against Symfony recommendations, as such this base class is already **deprecated** and will be removed in PrestaShop 10.0.
```
```suggestion
To avoid a very big breaking change, we modified the `PrestaShopBundle\Controller\Admin\FrameworkBundleAdminController`, so that it can fetch services from the controller's container **and** from the global one, but this is a little hack that goes against Symfony recommendations, as such this base class is already **deprecated and will be removed in PrestaShop 10.0**.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+
+This gives developer some time to refacto their controller, they should now rely on the new `PrestaShopBundle\Controller\Admin\PrestaShopAdminController` which doesn't implement this hack in order to force a proper implementation of the controllers.
```
```suggestion
This should give developers some time to refacto their controllers. They should now rely on the new `PrestaShopBundle\Controller\Admin\PrestaShopAdminController`, which doesn't implement the mentioned hack to force a proper implementation of the controllers.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+You can read more details about <a href="https://symfony.com/doc/6.4/controller/service.html" target="_blank">Controllers as services</a>, <a href="https://symfony.com/doc/6.4/service_container/service_subscribers_locators.html" target="_blank">Service Subscribers & Locators</a> or <a href="https://symfony.com/doc/6.4/service_container.html">Service container</a> in general.
+
+As an example you have three main possibilities to injected a service into your controller:
```
```suggestion
As an example, you have three main possibilities to inject a service into your controller:
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+
+```php
+<?php
+
+use PrestaShopBundle\Controller\Admin\PrestaShopAdminController;
+
+class MyController extends PrestaShopAdminController {
```
```suggestion
class MyController extends PrestaShopAdminController
{
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+     * 
+     * @param MyCustomService $myCustomService
+     */
+    public function __construct(
+      private readonly MyCustomService $myCustomService,
+    ) {
+    }
+    
+    /**
+     * You can inject a service only on one method, useful if the service is used only in one action, it is more optimized than injecting it all the time and never use it.
```
```suggestion
     * You can inject a service only on one method, useful if the service is used only in one action. It is more optimized than injecting it all the time and never using it.
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+        $specificData = $specifiService->getSpecificData();
+
+        // You can use the get method to fetch a service that was added in the getSubscribedServices method
+        $toolingService = $this->container->get(ToolingService::class);
+        
+        return new JsonResponse($toolingService->format($generalData + $specificData));
+    }
+    
+    /**
+     * You can define a list of registered services, they will be accessible via the $this->>container->get method, this
```
```suggestion
     * You can define a list of registered services, they will be accessible via the $this->container->get method, this
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+     * way of injecting service is interesting if implemented on a base controller class for generic services that
```
kpodemski, line 128: I didn't know about it, it's indeed interesting 👀

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+     * may be used by many controllers (like the router, the translator, ...) so you don't need to inject them manually
+     * in each controller class.
+     * 
+     * @return array
+     */
+    public static function getSubscribedServices(): array
+    {
+        return return parent::getSubscribedServices() + [
```
```suggestion
        return parent::getSubscribedServices() + [
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+You can already anticipate and define your controllers as services, it is also compatible with previous versions of Symfony and PrestaShop:
+
+```yaml
+services:
+  MyNamespace\MyController:
+    # Using autowire will inject services in the constructors automatically
+    autowire: true
+    # Using autoconfigure with controllers that extend AbstractController will handle method parameter injection and subscribed services
+    autoconfigure: true
+    # If your controller class doesn't extend AbstractController you should add this tag manually si you benefit from method parameters injection
```
```suggestion
    # If your controller class doesn't extend AbstractController, you should add this tag manually, so you benefit from method parameters injection
```

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+    tags: ['controller.service_arguments']
+```
+
+{{% notice note %}}
+**PrestaShopAdminController::getSubscribedServices**
+
+The new base controller class uses the `getSubscribedServices` method to give easy access to some commonly used PrestaShop services, like the `ConfigurationInterface`, `HookDispatcherInterface`, `TranslatorInterface`, ...
+You can see the <a href="https://github.com/PrestaShop/PrestaShop/blob/develop/src/PrestaShopBundle/Controller/Admin/PrestaShopAdminController.php#L59" target="_blank">full list of subscribed services</a> in the class, this list will probably grow piece by piece but only for common generic services.
```
```suggestion
You can see the <a href="https://github.com/PrestaShop/PrestaShop/blob/9.0.0-alpha.1/src/PrestaShopBundle/Controller/Admin/PrestaShopAdminController.php#L59" target="_blank">full list of subscribed services</a> in the class, this list will probably grow piece by piece but only for common generic services.
```

we'll change to v9 final once ready

modules/core-updates/9.0.md
```diff
@@ -78,7 +78,91 @@ Some dependencies are still present but were upgraded which comes with their own
...
+<br/><br/>
+Thanks to this you can use `$this->container->get(ConfigurationInterface::class)->get('my_config')` to access the configuration service, or alternatively use the helper method `$this->getConfiguration()->get('my_config')`.
```
```suggestion
Thanks to this, you can use `$this->container->get(ConfigurationInterface::class)->get('my_config')` to access the configuration service, or alternatively use the helper method `$this->getConfiguration()->get('my_config')`.
```

modules/core-updates/9.0.md
```diff
@@ -178,6 +262,29 @@ If you need to integrate your code with the authentication process you should no
 - [AdminController::isAnonymousAllowed is now public](https://github.com/PrestaShop/PrestaShop/pull/35810)
 - [Login page and authentication migrated](https://github.com/PrestaShop/PrestaShop/pull/35983)
 
+### Symfony context refacto
+
+Lots of core code is based on the legacy `Context` class that stores some data as a singleton, but this legacy implementation has many drawbacks so a refactoring was started in PrestaShop 9.0 to replace this usage.
```
```suggestion
A significant amount of code in the core relies on the legacy 'Context' class, which stores certain data as a singleton. However, this legacy implementation has many drawbacks, so refactoring was initiated in PrestaShop 9.0 to replace this usage.
```

modules/core-updates/9.0.md
```diff
@@ -178,6 +262,29 @@ If you need to integrate your code with the authentication process you should no
...
+
+- `PrestaShop\PrestaShop\Core\Context\ApiClientContext`
+- `PrestaShop\PrestaShop\Core\Context\CountryContext`
+- `PrestaShop\PrestaShop\Core\Context\CurrencyContext`
+- `PrestaShop\PrestaShop\Core\Context\EmployeeContext`
+- `PrestaShop\PrestaShop\Core\Context\LanguageContext`
+- `PrestaShop\PrestaShop\Core\Context\LegacyControllerContext`
+- `PrestaShop\PrestaShop\Core\Context\ShopContext`
+
+We still have some work to do to replace all the usages of the legacy context with the modern ones, but new code will stop using the legacy ones. We recommend developers to rely on the new ones from now on.
```
```suggestion
We still have some work to do to replace all the usages of the legacy context with the modern ones, but the new code will stop using the legacy ones. We recommend that developers rely on the new ones from now on.
```

kpodemski (commented):
Some suggestions.

I think you touched on some fundamental subjects that should go into the general documentation.

modules/core-updates/9.0.md
jolelievre: Yep it's quite elegant, it's used by the framework without us realizing it

modules/core-updates/9.0.md
jolelievre: Yeah that's what I though as well, that's why I used the branch for now But okay for that tag

modules/core-updates/9.0.md
jolelievre: Ok except for `Context` that should be in backquotes IMHO, as it's a reference to some code

jolelievre:
Thanks for the review @kpodemski 
I addressed all your feedback I agree that this page defines many important changes that should probably be mentioned in other places As I proposed when I started documenting these changes, the idea is to do it all here for now, just because it was more convenient for me

But once everything is documented (only this PR and the one about Symfony layout left), we should indeed think about how to improve it:
- some parts deserve a dedicated page
- some parts should be copied in other relevant places probably

kpodemski (approved):

PR #1829, opened 2024-06-19
Detail the impacts of the new Symfony layout

jolelievre:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Detail the impacts of the new Symfony layout
| Fixed ticket? | ~

### DIFF
--- a/modules/core-updates/9.0.md
+++ b/modules/core-updates/9.0.md
@@ -202,6 +202,97 @@ Display event listeners for the back office in `dev` environment (default value 
 - [Experimental front container](https://github.com/PrestaShop/PrestaShop/pull/32719)
 - [Admin API has dedicated kernel, we now have three distinct configurations](https://github.com/PrestaShop/PrestaShop/pull/35515)
 
+### Symfony layout
+
+All the back office pages share a common layout, composed of a few elements:
+- the `<head>` element that includes all the CSS and JS (among other things)
+- the side navigation menu
+- the header of the page which is itself composed of:
+  - quick accesses
+  - search form
+  - notifications center
+  - employee dropdown
+  - the multistore header (when multistore is enabled)
+  - the page toolbar (which includes breadcrumb and top action buttons)
+- the footer (which only contains a displayBackOfficeFooter display hook)
+
+Until PrestaShop 8.1, all these common elements were handled by legacy code, so even on the migrated pages, there was always a background legacy controller based on `AdminController` in charge of building the layout data and rendering it. Once the layout was rendered the central content of Symfony pages was included in the middle of it. It means no page was completely free from the legacy controllers, which would ultimately block the end of migration, while they are intended to disappear completely.
+
+In PrestaShop 9.0 all this layout part is now fully handled by Symfony, we use <a href="https://symfony.com/bundles/ux-twig-component/current/index.html" target="_blank">Twig components</a> to render each element independently. The code is, therefore, easier to understand, a component class is responsible for fetching/building the data, while the actual rendering is based on Twig. See our <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component" target="_blank">Layout components</a> for more details.
+
+On legacy pages, we follow the same principle. Symfony is now in charge of rendering all the layout, and we use some <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component/Legacy" target="_blank">Legacy layout components</a> that follow the same architecture, but render different Twig templates to fit with the old **default theme** (that is based on old Bootstrap and includes old legacy helpers like `HelperForm` and `HelperList`).
+
+**What changes for my module pages?**
+
+We refactored this layout with maximum backward compatibility in mind. The HTML layout itself had minimum changes (in both migrated and legacy pages). We tried to keep the same hooks in both the PHP code and the Twig templates. We even introduced a fake legacy controller in migrated pages that has no logic but is kept mostly as a [DTO](https://en.wikipedia.org/wiki/Data_transfer_object) to contain things like CSS files and JS files, because that's where most modules add their content.
+We still had to change the controller workflow. Many of the functions in `AdminController` are no longer useful as they are used to initialize the layout variables (now handled by Twig components). Some methods also render or write output directly, which we prevent as we need to get the rendered content as a string to integrate it correctly inside the new layout. This means some internal methods had to be split to avoid unwanted usage.
+
+**Legacy workflow:**
+
+- `Dispatcher::dispatch` would detect the legacy controller and instantiate it
+- `AdminController::run` was executed by the `Dispatcher`, which was split into
+  - `AdminController::init`
+  - `AdminController::checkAccess` would handle the check of the legacy token in the url
+  - `AdminController::setMedia` is in charge of defining the JS/CSS files to be loaded in the page
+  - `AdminController::postProcess` processes the posted data
+    - if after this method `AdminController::redirect_after` is defined then perform an HTTP redirection
+  - `AdminController::initHeader`
+  - `AdminController::initContent`
+  - `AdminController::initFooter`
+  - Check if `AdminController::ajax`, one of the two is executed
+    - `AdminController::displayAjax` (or specific `displayAjax{action}` method)
+    - `AdminController::display`
+      - init many variables used by the layout templates (meta title, toolbar, JS definitions, ...)
+      - render `page`, `header`, `footer` and sets them as Smarty variables
+      - fetch the Smarty `AdminController::template` to get the content of the page
+      - calls `smartyOutputContent` to echo/output the whole page (layout + content)
+
+Here you'll find the [execution workflow of legacy controllers]({{< relref "9/development/architecture/legacy/legacy-controllers.md" >}})
+
+**Symfony workflow:**
+
+- A special routing matching URLs like `?controller=LegacyAdminController` checks if the URL is valid thanks to `LegacyRouterChecker::check`
+  - the query parameter must match an existing legacy controller class
+  - the controller object is instantiated and `AdminController::init` is called, this is important to know if the controller is anonymous, this is the proper moment to define `AdminController::allowAnonymous`
+  - a few attributes are set on the Symfony `Request` so they can be reused later in the workflow by other components/services
+- `LegacyController::legacyPageAction` is a generic controller responsible for executing the legacy controller, it doesn't execute `AdminController::run` but splits it into several steps
+  - `AdminController::checkAccess` is not called, the token in URLs is a Symfony CSRF token now and is checked by `TokenizedUrlListener`
+  - `LegacyController::initController`
+    - the permission is checked based on the detected controller and action, unless `AdminController::viewAccess` has been overridden in which case it is used to check authorization
+    - `SmartyVariablesFiller::fillDefault` is in charge of building most of the generic smarty variables previously handled by `AdminController::initHeader`
+    - `AdminController::setMedia` is in charge of defining the JS/CSS files to be loaded in the page
+    - `AdminController::postProcess` processes the posted data
+      - if after this method `AdminController::redirect_after` is defined then perform an HTTP redirection
+  - Check if `ajax` query parameter is defined, one of the two is executed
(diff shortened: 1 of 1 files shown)


modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+- the `<head>` element that includes all the CSS and JS (among other things)
+- the side navigation menu
+- the header of the page which is itself composed of
+- quick accesses
+- search form
+- notifications center
+- employee dropdown
+- the multistore header (when multistore is enabled)
+- the page toolbar (which includes breadcrumb and top action buttons)
+- the footer (which only contains a displayBackOfficeFooter display hook)
```
```suggestion
All the back office pages share a common layout, composed of a few elements:
- the `<head>` element that includes all the CSS and JS (among other things)
- the side navigation menu
- the header of the page, which is itself composed of:
-- quick accesses
-- search form
-- notifications center
-- employee dropdown
-- the multistore header (when multistore is enabled)
-- the page toolbar (which includes breadcrumb and top action buttons)
- the footer (which only contains a `displayBackOfficeFooter` hook)
```

formatting

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+Until PrestaShop 8.1 all these common elements were handled by legacy code, so even on migrated page there was always a background legacy controller based on `AdminController` in charge of building the layout data and rendering it. Once the layout was rendered the central content of Symfony pages was included in the middle of it. It means no page was completely free from the legacy controllers which would ultimately block the end of migration where they are meant to disappear completely.
```
```suggestion
Until PrestaShop 8.1, all these common elements were handled by legacy code, so even on the migrated page, there was always a background legacy controller based on `AdminController` in charge of building the layout data and rendering it. Once the layout was rendered the central content of Symfony pages was included in the middle of it. It means no page was completely free from the legacy controllers, which would ultimately block the end of migration, while they are intended to disappear completely.
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+In PrestaShop 9.0 all this layout part is now fully handled by Symfony, we use <a href="https://symfony.com/bundles/ux-twig-component/current/index.html" target="_blank">Twig components</a> to render each element independently. The code is, therefore, easier to understand, a component class is responsible for fetching/building the data while the actual rendering is based on Twig. See our <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component" target="_blank">Layout components</a> for more details.
```
```suggestion
In PrestaShop 9.0 all this layout part is now fully handled by Symfony, we use <a href="https://symfony.com/bundles/ux-twig-component/current/index.html" target="_blank">Twig components</a> to render each element independently. The code is, therefore, easier to understand, a component class is responsible for fetching/building the data, while the actual rendering is based on Twig. See our <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component" target="_blank">Layout components</a> for more details.
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+In legacy pages the same principle is used, Symfony is now in charge of rendering all the layout, and we use some <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component/Legacy" target="_blank">Legacy layout components</a> that follow the same architecture but render different twig templates to fit with the old **default theme** (that is based on old bootstrap and includes old legacy helpers like form and lists).
```
```suggestion
On legacy pages, we follow the same principle. Symfony is now in charge of rendering all the layout, and we use some <a href="https://github.com/PrestaShop/PrestaShop/tree/develop/src/PrestaShopBundle/Twig/Component/Legacy" target="_blank">Legacy layout components</a> that follow the same architecture, but render different Twig templates to fit with the old **default theme** (that is based on old Bootstrap and includes old legacy helpers like `HelperForm` and `HelperList`).
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+**So what changes for my module pages?**
```
```suggestion
**What changes for my module pages?**
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+We refactored this layout with maximum backward compatibility in mind, the HTML layout itself had minimum changes (in both migrated and legacy pages), we tried to keep the same hooks in both the PHP code and the twig templates. We even introduced a fake legacy controller in migrated pages that has no logic but is kept mostly as a DTO to contain things like CSS files and JS files, because that's where most modules add their content.
```
```suggestion
We refactored this layout with maximum backward compatibility in mind. The HTML layout itself had minimum changes (in both migrated and legacy pages). We tried to keep the same hooks in both the PHP code and the Twig templates. We even introduced a fake legacy controller in migrated pages that has no logic but is kept mostly as a [DTO](https://en.wikipedia.org/wiki/Data_transfer_object) to contain things like CSS files and JS files because that's where most modules add their content.
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+But we still had to change the controller workflow, many code in `AdminController` is now not useful anymore as it is used to initialize the layout variables (now handled by Twig components) some methods also renders or write output directly which we prevent as we need to get the rendered content as a string to correctly integrate it inside the new layout so some internal methods had to be split to avoid unwanted usages.
```
```suggestion
We still had to change the controller workflow. Many of the functions in `AdminController` are no longer useful as they are used to initialize the layout variables (now handled by Twig components). Some methods also render or write output directly, which we prevent as we need to get the rendered content as a string to integrate it correctly inside the new layout. This means some internal methods had to be split to avoid unwanted usage.
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+
+**Legacy workflow:**
+
+- `Dispatcher::dispatch` would detect the legacy controller and instanciate it
```
```suggestion
- `Dispatcher::dispatch` would detect the legacy controller and instantiate it
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+- `AdminController::run` was executed by the `Dispatcher` which was split into
```
```suggestion
- `AdminController::run` was executed by the `Dispatcher`, which was split into
```

modules/core-updates/9.0.md
kpodemski: either legacy workflow is still somewhere, and we should keep "detects" or it's a song of the past, and we should use `would detect`, this is relevant for all other elements on this list

kpodemski: maybe you could link to this page https://devdocs.prestashop-project.org/8/development/architecture/legacy/legacy-controllers/?

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+      - init many variables used by the layout templates (meta title, toolbar, JS definitions, ...)
+      - render `page`, `header`, `footer` and sets them as Smarty variables
+      - fetch the Smarty `AdminController::template` to get the content of the page
+      - calls `smartyOutputContent` to echo/output the whole page (layout + content)
+
+**Symfony workflow:**
+
+- A special routing matching URLs like `?controller=LegacyAdminController` checks if the URL is valid thanks to `LegacyRouterChecker::check`
+  - the query parameter must match an existing legacy controller class
+  - the controller object is instanciated and `AdminController::init` is called, this is important to know if the controller is anonymous, this is the proper moment to define `AdminController::allowAnonymous`
```
```suggestion
  - the controller object is instantiated, and `AdminController::init` is called, this is important to know if the controller is anonymous, this is the proper moment to define `AdminController::allowAnonymous`
```

kpodemski: consider explaining what it means to call a controller "anonymous"

kpodemski: also, we use Mermaid in the docs, so if you want, you could use https://mermaid.live/edit to generate flowchart :D

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+      - if after this method `AdminController::redirect_after` is defined then perform an HTTP redirection
+  - Check if `ajax` query parameter is defined, one of the two is executed
+    - `LegacyController::renderAjaxController` executes `AdminController::initContent` the appropriate `AdminController::displayAjax` method, it catches all the output data into a string buffer (because most legacy ajax methods simply echo their content) and returns it as a response
+    - `LegacyController::renderPageContent`
+      - executes `AdminController::initContent`
+      - fetch the Smarty `AdminController::template` to get the content of the page
+      - calls `Cookie::write` which was previously called via `smartyOutputContent`
+      - get the modal content via `AdminController::renderModal`
+
+The two workflows should render the same result, many methods from the legacy workflow are not executed anymore because they lost their purpose, but in case you overrode one of those methods be aware that they are no longer called directly so they are breaking changes:
```
```suggestion
The two workflows should render the same result. Many methods from the legacy workflow are not executed anymore because they lost their purpose, but in case you override one of those methods, be aware that they are no longer called directly, so there are breaking changes:
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+- `AdminController:run`
+- `AdminController::initHeader`
+- `AdminController::initContent`
+- `AdminController::initFooter`
+- `AdminController:display`
+
+Also be aware that many smarty variables are no longer defined because they were only used to render the layout.
```
```suggestion
Also, be aware that many Smarty variables are no longer defined because they were only used to render the layout.
```

modules/core-updates/9.0.md
```diff
@@ -117,6 +117,95 @@ Display event listeners for the back office in `dev` environment (default value
...
+- `actionDispatcherBefore`
+- `actionDispatcherAfter`
+
+**Related PRs:**
+- [Smarty variables viewport_scale and inline_js variables no longer present](https://github.com/PrestaShop/PrestaShop/pull/33775)
+- [Removed modal_module_list, modals views in new layout](https://github.com/PrestaShop/PrestaShop/pull/33864)
+- [Clean Smarty variables that were only relevant for the layout internal use](https://github.com/PrestaShop/PrestaShop/pull/34560)
+- [Legacy pages are rendered by Symfony](https://github.com/PrestaShop/PrestaShop/pull/34783)
+
+For more details about the changes you can check the content of the <a href="https://github.com/PrestaShop/PrestaShop/issues/32875" target="_blank">Symfony layout Epic</a>.
```
```suggestion
For more details about the changes, you can check the content of the <a href="https://github.com/PrestaShop/PrestaShop/issues/32875" target="_blank">Symfony layout Epic</a>.
```

kpodemski (commented):
I added some suggestions.

If you have some time, consider using Mermaid, especially since a live editor that generates the code is available. If not, I'll try to find some time to improve this page afterward.

On my side, I will test some of the legacy modules I have to better understand the impact of these changes.

Thanks @jolelievre 🙏🏻

modules/core-updates/9.0.md
jolelievre: It's in the past for the dispatcher, for other methods some are still called So I tried to keep some consistency and use present for methods that are still used

I didn't know about this diagram, it's great!! We should indeed create a similar one for the new layout I didn't know about mermaid but it seems great!

I'd prefer if we add this workflow in another PR though, because I'm on vacation at the end of the week and I'd like to finalize all the text content this week at least.

modules/core-updates/9.0.md
jolelievre: Yep good idea, but as I said in another comment probably in another PR

jolelievre:
Thanks @kpodemski 

I think I addressed all your feedback, I agree that there's still room for improvement Especially adding a diagram for the Symfony workflow, but I'd prefer to do it in another PR.

I'm leaving for three weeks and I won't have time to finish it this week.

kpodemski (approved):
@jolelievre sure thing! enjoy your holidays!

PR #1831, opened 2024-06-25
Update misleading create-custom-column-type.md

krzysztofgal:
Change column type name and twig template name. Using **html.html.twig** overrides this template on **all pages** so with "html" this tutorial is misleading.

| Questions     | Answers
| Branch?       | 8.x
| Description?  | I followed this tutorial and had very unexpected result on PS 8.1.7. This tutorial overrides template instead of creating new custom column.
| Fixed ticket? | -


kpodemski (approved):

kpodemski:
thanks @krzysztofgal - it definitely makes sense

PR #1855, opened 2024-08-06
Recommend PHP 8.3 by default

Hlavtox:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | PHP 8.1 will go outdated in few months, so we should recommend PHP 8.3 by default for PrestaShop 9. All modules must be 100% compatible with PHP 8.3, if they want to claim compatibility with Prestashop 9. There is no reason NOT to recommend the latest version of PHP.
| Fixed ticket? |


kpodemski (approved):

ShaiMagal:
I'm happy, that it's going this way.

PR #1862, opened 2024-09-13
Add bootstrap compatibility layer page

jolelievre:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Add page explaining the compatibility layer for bootstrap
| Fixed ticket? | ~

### DIFF
--- a/themes/reference/bootstrap-compatibility.md
+++ b/themes/reference/bootstrap-compatibility.md
@@ -1,0 +1,93 @@
+---
+title: Bootstrap compatibility
+---
+
+# Bootstrap compatibility
+
+Starting from PrestaShop 9.0 you can choose between two themes `classic` or `hummingbird`, the Hummingbird theme is based on more modern design and dependencies and gives better performance and SEO.
+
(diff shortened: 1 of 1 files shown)


Hlavtox (commented):
I don't think it's a good idea to promote the compatibility layer. If somebody goes through making his module compatible, he should make it properly…

It's not that much work to change few classes here and there and make a proper TPL for Bootstrap 5. Also the javascript is easy to migrate.

ga-devfront (approved):

ga-devfront:
> I don't think it's a good idea to promote the compatibility layer. If somebody goes through making his module compatible, he should make it properly…
> 
> It's not that much work to change few classes here and there and make a proper TPL for Bootstrap 5. Also the javascript is easy to migrate.

I agree but it is also our duty to support and offer a quick solution. Some modules/themes are very large and changing classes/js can take a lot of time, which is time when the module/theme may no longer sell.
I think it is very good to explain the solution while emphasizing that it is still preferable to update the module/theme.

matks (approved):
Agreed with @ga-devfront : we simply document the existence of the library.

I see it as a polyfill. It’s here to help people with complex projects.

If we did not document things that could be used wrong, should we remove the override documentation 😉 ?

PR #1867, opened 2024-09-25
Update upgrade.md

sebastian-marinescu:
- Add newest 8.x version as example
- Better unzip and copy example
- Remove install/module from list of folders to delete (not sure, but later in the text, it seems as if "install" is actually needed...)

Branch: 8.x


kpodemski (approved):

kpodemski:
thank you @sebastian-marinescu

PR #1875, opened 2024-10-29
Update httpd.md to include FilesMatch example for PHP-FPM in other container with different mount path

pavonic:
| Questions     | Answers
| Branch?       | 8.x 

### DIFF
--- a/basics/installation/advanced/httpd.md
+++ b/basics/installation/advanced/httpd.md
@@ -114,9 +114,7 @@ Don't forget to edit this configuration to make it works.
 </VirtualHost>
 ```
 
-If PHP-FPM is on another container than APACHE and both containers share a single volume that contains
-the application code but with different mount path, then the use and adapt the following lines (in this
-example /var/www/html is the mount path in the PHP-FPM container)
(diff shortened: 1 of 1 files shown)


basics/installation/advanced/httpd.md
```diff
@@ -113,3 +113,15 @@ Don't forget to edit this configuration to make it works.
     CustomLog /var/log/apache2/prestashop.access.log combined
 </VirtualHost>
 ```
+
+If PHP-FPM is on another container than APACHE and both containers share a single volume that contains
+the application code but with different mount path, then the use and adapt the following lines (in this
+example /var/www/html is the mount path in the PHP-FPM container)
```
kpodemski:
```suggestion
If PHP-FPM is running in a separate container from Apache, and both containers share a single volume that contains the application code but with different mount paths, you should use and adapt the code snippet below. In the example below, the mount path in the PHP-FPM container is /var/www/html.
```

kpodemski (approved):

kpodemski:
thank you @pavonic 

I rephrased it a bit, but the rest is all good, thank you for contributing to the docs!

PR #1893, opened 2024-11-28
Remove "skip_backup" option from CLI 

PierreEtSol:
It seems the option "skip_backup" is removed from CLI and replaced by BO config value. https://github.com/PrestaShop/autoupgrade/pull/814/files/142b3d7f0467b1834da87b95a57ba7ea216e1ba5

Stores with huge file number (unaware of BO config and only want CLI) are maybe mislead.

### DIFF
--- a/basics/keeping-up-to-date/upgrade-module/upgrade-cli.md
+++ b/basics/keeping-up-to-date/upgrade-module/upgrade-cli.md
@@ -94,7 +94,7 @@ To upgrade your PrestaShop store to the latest version using the command line in
 
 1. Create a configuration file for the AutoUpgrade module to use the local archive. Adjust the settings as needed:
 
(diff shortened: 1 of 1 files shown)


kpodemski:
Hello @PierreEtSol 

I think that to make things clear, we should use `PS_AUTOUP_BACKUP` config name instead, correct?

PierreEtSol:
> Hello @PierreEtSol
> 
> I think that to make things clear, we should use `PS_AUTOUP_BACKUP` config name instead, correct?

I think we just need to remove this parameter from CLI because it is not used anymore.
The command internally relies now on  `PS_AUTOUP_BACKUP`

Quetzacoalt91:
With the release of Update Assistant v7, the backup is now managed with a dedicated command, updates won't create them anymore.
See https://devdocs.prestashop-project.org/8/basics/keeping-up-to-date/update/update-from-the-cli/#backupcreate-command to discover how to create the backups.

Hopefully this makes the use of the module easier, and we can close this pull-request as it deals with a removed parameter.

PR #1895, opened 2024-12-01
Fix Typo in phpMyAdmin example

randallmorse:
the service is mysql not some-mysql

| Questions     | Answers
| Branch?       |  8.x 
| Description?  | The docker compose example for adding phpMyAdmin has a typo in its links definition, the example fails as its looking for the service name instead of the container name, additionally ensure the the container can reach the mysql service by adding it to the shared private network.

### DIFF
--- a/basics/installation/environments/docker.md
+++ b/basics/installation/environments/docker.md
@@ -344,6 +344,8 @@ services:
         PMA_HOST: some-mysql
         PMA_PORT: 3306
         PMA_ARBITRARY: 1
+        PMA_USER: root
+        PMA_PASSWORD: admin
       restart: unless-stopped
       ports:
         - 8081:80
--- a/basics/installation/environments/docker.md
+++ b/basics/installation/environments/docker.md
@@ -347,6 +347,8 @@ services:
       restart: unless-stopped
       ports:
         - 8081:80
+      networks:
+        - prestashop_network
 ```
 
 And access `http://localhost:8081` to access `phpMyAdmin`.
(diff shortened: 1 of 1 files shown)


basics/installation/environments/docker.md
```diff
@@ -339,7 +339,7 @@ services:
       image: phpmyadmin/phpmyadmin
       container_name: phpmyadmin
       links:
-        - some-mysql
+        - mysql
       environment:
         PMA_HOST: some-mysql
```
SharakPL, line 344:
```suggestion
        PMA_HOST: mysql
```
This also should be changed to `mysql` as the host ~~must match~~ the database service name too.

Edit: actually it looks like it works both ways. Either service name or container name will link phpmyadmin properly.

SharakPL (changes requested):

SharakPL (approved):

basics/installation/environments/docker.md
```diff
@@ -339,14 +339,18 @@ services:
...
         PMA_PORT: 3306
         PMA_ARBITRARY: 1
+        PMA_USER: root
+        PMA_PASSWORD: admin
```
randallmorse, line 348: @SharakPL is this the intent of having the host name and port set here? see comment on the commit, at current it isnt used unless the the PMA_HOST and PMA_PASSWORD is set, I think this might have been the intent and if not could be a great add to the example to assist new users.

kpodemski (approved):

kpodemski:
thank you @randallmorse

PR #1902, opened 2024-12-19
Regenerate hooks list documentations

### LINKED ISSUE #37163: Enhance the hook list page
Currently, [this list](https://devdocs.prestashop-project.org/8/modules/concepts/hooks/list-of-hooks/) lacks information to enable new contributors to research and understand how existing hooks work, and therefore to avoid duplication with existing ones

1. Enhance template for hook page
2. Use the script to update each hook page (ensure that every hooks from the file `hook.xml`  are exported)
3. Add the following columns in the list : 
  - a short description
  - the type (display, action)
4. Remove the column `Alias`

tleon:
| Questions     | Answers
| Branch?       | 8.x
| Description?  | Regenerate hooks list documentations
| Fixed ticket? | Fixes #37163

This was generated using the command from the core with:
```sh
./bin/console prestashop:update:hooks-documentation /Users/myUser/devdocs-site/src/content/9/modules/concepts/hooks/list-of-hooks
```

### DIFF
--- a/modules/concepts/hooks/list-of-hooks/actionAddressGridDefinitionModifier.md
+++ b/modules/concepts/hooks/list-of-hooks/actionAddressGridDefinitionModifier.md
@@ -1,0 +1,29 @@
+---
+Title: actionAddressGridDefinitionModifier
+hidden: true
+hookTitle: 'Modify address grid definition'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/Core/Grid/Definition/Factory/AbstractGridDefinitionFactory.php'
+        file: src/Core/Grid/Definition/Factory/AbstractGridDefinitionFactory.php
+locations:
+    - 'back office'
+type: action
+hookAliases: 
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook allows to alter address grid columns, actions and filters'
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+$this->hookDispatcher->dispatchWithParameters('action' . Container::camelize($definition->getId()) . 'GridDefinitionModifier', [
+    'definition' => $definition,
+]);
+```
--- a/modules/concepts/hooks/list-of-hooks/displayBackOfficeFooter.md
+++ b/modules/concepts/hooks/list-of-hooks/displayBackOfficeFooter.md
@@ -1,0 +1,27 @@
+---
+Title: displayBackOfficeFooter
+hidden: true
+hookTitle: 'Administration panel footer'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/PrestaShopBundle/Resources/views/Admin/Component/Layout/footer.html.twig'
+        file: src/PrestaShopBundle/Resources/views/Admin/Component/Layout/footer.html.twig
+locations:
+    - 'back office'
+type: display
+hookAliases: displayBackOfficeFooter
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook is displayed within the admin panel''s footer'
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+{{ renderhook('displayBackOfficeFooter') }};
+```
--- a/modules/concepts/hooks/list-of-hooks/actionAddressGridDataModifier.md
+++ b/modules/concepts/hooks/list-of-hooks/actionAddressGridDataModifier.md
@@ -1,0 +1,29 @@
+---
+Title: actionAddressGridDataModifier
+hidden: true
+hookTitle: 'Modify address grid data'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/Core/Grid/GridFactory.php'
+        file: src/Core/Grid/GridFactory.php
+locations:
+    - 'back office'
+type: action
+hookAliases: 
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook allows to modify address grid data'
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+$this->hookDispatcher->dispatchWithParameters('action' . Container::camelize($definition->getId()) . 'GridDataModifier', [
+    'data' => &$data,
+]);
+```
--- a/modules/concepts/hooks/list-of-hooks/actionAddressGridFilterFormModifier.md
+++ b/modules/concepts/hooks/list-of-hooks/actionAddressGridFilterFormModifier.md
@@ -1,0 +1,29 @@
+---
+Title: actionAddressGridFilterFormModifier
+hidden: true
+hookTitle: 'Modify address grid filters'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/Core/Grid/Filter/GridFilterFormFactory.php'
+        file: src/Core/Grid/Filter/GridFilterFormFactory.php
+locations:
+    - 'back office'
+type: action
+hookAliases: 
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook allows to modify filters for address grid'
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+$this->hookDispatcher->dispatchWithParameters('action' . Container::camelize($definition->getId()) . 'GridFilterFormModifier', [
+    'filter_form_builder' => $formBuilder,
+]);
+```
--- a/modules/concepts/hooks/list-of-hooks/actionAddressGridPresenterModifier.md
+++ b/modules/concepts/hooks/list-of-hooks/actionAddressGridPresenterModifier.md
@@ -1,0 +1,29 @@
+---
+Title: actionAddressGridPresenterModifier
+hidden: true
+hookTitle: 'Modify address grid template data'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/Core/Grid/Presenter/GridPresenter.php'
+        file: src/Core/Grid/Presenter/GridPresenter.php
+locations:
+    - 'back office'
+type: action
+hookAliases: 
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook allows to modify data which is about to be used in template for address grid'
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+$this->hookDispatcher->dispatchWithParameters('action' . Container::camelize($definition->getId()) . 'GridPresenterModifier', [
+    'presented_grid' => &$presentedGrid,
+]);
+```
--- a/modules/concepts/hooks/list-of-hooks/actionAddressGridQueryBuilderModifier.md
+++ b/modules/concepts/hooks/list-of-hooks/actionAddressGridQueryBuilderModifier.md
@@ -1,0 +1,31 @@
+---
+Title: actionAddressGridQueryBuilderModifier
+hidden: true
+hookTitle: 'Modify address grid query builder'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/src/Core/Grid/Data/Factory/DoctrineGridDataFactory.php'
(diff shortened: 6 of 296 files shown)


nicosomb:
Why on 8.0.x branch?

jolelievre (changes requested):
I think the command has room for improvements:
- it is removing some valid documentation like `action<ClassName><Action>Before.md`, these hooks were detected but skipped in the command so I guess that's why they are removed here
- we didn't have infos extracted in the command bu here we have some data in the md files, I think the generation command should be able to also parse the data from the documentation from the md with regexp This way if we are not able to generate the description or usages they can be manually added in the doc Then on next generation the command can extract these data and it would avoid erasing some valid content

I block the PR for now until we improve the command but it's a really good start it just need some updates once the command has been improved

Also as @nicosomb said is the target here 8.X on purpose? does it mean you ran this command from the 8.1.x branch? it seems unlikely since the command has only been add on 9.0.x so far so maybe it's the wonrg base branch

kpodemski (changes requested):
I'm blocking it because there are too many issues at the moment.

One thing you probably weren't aware of:
You need to implement a feature in the generator to let us ignore some of the hooks; some from the docs have additional information, and some examples have been added that cannot be generated.

Hlavtox:
Same opinion as others - there is many useful information in the docs that have been contributed, it would get lost.

tleon:
So I targeted the right branch and the right files. 

Regarding the data loss : 

I ran the doc generation command directly on the current version of the md folder ( previously I deleted all md in there and regenerated all of them. I was not aware that some of the md may have been modified by hand.) 

I'm afraid that even while running the command like this, the files already there have been erased as  before. 

What I could do is : modify the command to ignore the files already present in the folder ( matching on the name of the files ). This would preserve the md modified by hand but those ones would miss some informations that I added in the command. ( I think ;) ) 

Also Like it was suggested I can ignore the hook containing a placeholder in the name I did it in the generation of the xml file I can just do the same thing here.

How does that sound ? 

@kpodemski @Hlavtox @nicosomb @jolelievre

jolelievre:
Hi @tleon 

ok good to know that the removal was manual and not caused by the command. I think we can think later about how we could remove automatically the obsolete hooks but let's not get carried away for now.

Regarding the override of existing hooks your suggestion is interesting but I'm thinking we could go one step further, the idea would be that if the file is already present the command only updates the fields that are missing or empty. Any already existing data/text in the MD files would be kept by default so only new data would be used for override.

To implement this we could add a `HookDocExtractor` service that can read and parse the data for each hook based on each MD file name. Then when you create the data to be exported the priority would be for each field something like `Doc > XML fixture > Auto generated data > Default fallback`


On the side note while you're on this topic some people on slack noticed that some hook doc pages are empty like this one https://devdocs.prestashop-project.org/9/modules/concepts/hooks/list-of-hooks/displaypdftemplate/ which matches the `displayPDF<Template>` hook, one theory is that there is a problem because of the charrets `<>` but it's worth looking into it, it may not be the (only) reason because this one is fine https://devdocs.prestashop-project.org/9/modules/concepts/hooks/list-of-hooks/actionclassnameactionbefore/

kpodemski:
> To implement this we could add a HookDocExtractor service that can read and parse the data for each hook based on each MD file name. Then when you create the data to be exported the priority would be for each field something like Doc > XML fixture > Auto generated data > Default fallback

This may require some changes in how we format information about a given hook, and it sounds a bit complicated, but if it's feasible, and that it could be trusted, then it would ease work on maintaining the list

jolelievre:
After discussion with @tleon we decided that for now we will do a simple modification in the command The command checks if the MD file already exists it simply skips it, so the command only adds new hooks

This way we are sure that we don't remove some manually modified files that are likely more up to date

If we want to improve this doc generation it will be done in another PR but we'll have to think about what we clearly want and if there's some real value to do it

jolelievre (approved):
Ok for me but I'd like to have @kpodemski opinion on this one

modules/concepts/hooks/list-of-hooks/actionAddressGridDefinitionModifier.md
```diff
@@ -0,0 +1,27 @@
+---
+Title: actionAddressGridDefinitionModifier
+hidden: true
+hookTitle: 'Modify address grid definition'
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/'
```
kpodemski: I think it could be interesting to have an origin of this hook, it fits the format of action<GridId>GridDefinitionModifier and we know where the hook is.

modules/concepts/hooks/list-of-hooks/actionAddressGridDefinitionModifier.md
```diff
@@ -0,0 +1,27 @@
...
+
+---
+
+{{% hookDescriptor %}}
+
+## Call of the Hook in the origin file
+
+```php
+;
+```
```
kpodemski: This looks like a mistake. Again, it would be good to have a source for this hook here.

modules/concepts/hooks/list-of-hooks/actionAddressGridDefinitionModifier.md
kpodemski: and if it is not possible or hard to achieve, then maybe we don't need to list all of them but improve the search mechanism further according to this information:

```
Search tip: Some hooks are generated dynamically, so their names are documented in a generic way.

For example, actionAdminCustomersFormModifier is documented as action<AdminControllerClassName>FormModifier. A regex based search has been implemented, and generic hooks should be matched. However, when you see a controller name or action in the hook name and you can’t find it, try searching for a part of the hook name, like FormModifier.
```

from what I see most of the hooks that were generated are action hooks with the dynamic names

modules/concepts/hooks/list-of-hooks/actionAdministrationPageUploadQuotaSave.md
```diff
@@ -0,0 +1,28 @@
...
+locations:
+    - 'back office'
+type: action
+hookAliases: 
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook allows to modify data of administration page upload quota options form after it was saved
+      '
```
kpodemski: Probably z `trim` missing

modules/concepts/hooks/list-of-hooks/fooHook.md
```diff
@@ -0,0 +1,27 @@
+---
+Title: fooHook
+hidden: true
+hookTitle: ''
+files:
+    -
+        url: 'https://github.com/PrestaShop/PrestaShop/blob/9.0.x/tests/Unit/Core/Hook/HookDispatcherTest.php'
```
kpodemski: `tests` directory should be excluded :)

modules/concepts/hooks/list-of-hooks/displayTopColumn.md
```diff
@@ -0,0 +1,27 @@
+---
+Title: displayTopColumn
```
kpodemski: It's odd that there is no origin of this hook, it doesn't look like a dynamic one

modules/concepts/hooks/list-of-hooks/displayProductPageDrawer.md
```diff
@@ -0,0 +1,27 @@
+---
+Title: displayProductPageDrawer
```
kpodemski: again, it seems there is an issue with getting a source of the display hooks

modules/concepts/hooks/list-of-hooks/displayProductListFunctionalButtons.md
```diff
@@ -0,0 +1,27 @@
+---
+Title: displayProductListFunctionalButtons
```
kpodemski: I checked and it seems there is a hook registered but there is no use of it anywhere in the code 🤔 
Not sure what to do about it. Maybe of it is a display hook and there is no use, we could indicate it somehow 🤔  or this could help us clean the hooks in the db

cc @jolelievre

modules/concepts/hooks/list-of-hooks/displayCarrierList.md
```diff
@@ -0,0 +1,27 @@
+---
```
kpodemski: I noticed that this is the hook that is part of the deprecated hooks, here's the full list:

```
/**
     * List of all deprecated hooks.
     *
     * @var array
     */
    protected static $deprecated_hooks = [
        // Back office
        'backOfficeFooter' => ['from' => '1.7.0.0'],
        'displayBackOfficeFooter' => ['from' => '1.7.0.0'],

        // Shipping step
        'displayCarrierList' => ['from' => '1.7.0.0'],
        'extraCarrier' => ['from' => '1.7.0.0'],

        // Payment step
        'hookBackBeforePayment' => ['from' => '1.7.0.0'],
        'hookDisplayBeforePayment' => ['from' => '1.7.0.0'],
        'hookOverrideTOSDisplay' => ['from' => '1.7.0.0'],

        // Product page
        'displayProductTabContent' => ['from' => '1.7.0.0'],
        'displayProductTab' => ['from' => '1.7.0.0'],

        // Order page
        'displayAdminOrderRight' => ['from' => '1.7.7.0'],
        'displayAdminOrderLeft' => ['from' => '1.7.7.0'],
        'displayAdminOrderTabOrder' => ['from' => '1.7.7.0'],
        'displayAdminOrderTabShip' => ['from' => '1.7.7.0'],
        'displayAdminOrderContentOrder' => ['from' => '1.7.7.0'],
        'displayAdminOrderContentShip' => ['from' => '1.7.7.0'],

        // Controller
        'actionGetProductPropertiesAfter' => ['from' => '1.7.8.0'],
    ];
    ```
I think it would be good to skip them in the generator.

File: modules/concepts/hooks/list-of-hooks/displayBackOfficeFooter.md [hunk 9]
```diff
@@ -0,0 +1,27 @@
...
+        file: admin-dev/themes/default/template/footer.tpl
+locations:
+    - 'back office'
+type: display
+hookAliases: displayBackOfficeFooter
+array_return: false
+check_exceptions: false
+chain: false
+origin: core
+description: 'This hook is displayed within the admin panel''s footer'
```
kpodemski, line 17:
```suggestion
description: 'This hook is displayed within the admin panel''s footer'
```

I think this double `'` requires an escape

kpodemski (changes requested):
I added a few comments, there are some critical issues. I am also not sure if generating all the action hooks is a good idea... It makes the list huge, but if we could at least solve the problem with origin call and url to the file then it could help a lot.

modules/concepts/hooks/list-of-hooks/displayTopColumn.md
jolelievre: Seems like it was related to `homeslider` maybe an old module for 1.6

But the hook itself is not in the core anymore, this one is a remnant from the hook.xml file not a hook found by the extractor

modules/concepts/hooks/list-of-hooks/displayProductPageDrawer.md
jolelievre: SAme it's only in the hook.xml file

modules/concepts/hooks/list-of-hooks/displayProductListFunctionalButtons.md
jolelievre: Same only found in some XMLs and in example module's code I wonder if we should remove from the hook.xml all the hooks that are no longer found in the code base (except for dynamic ones)

modules/concepts/hooks/list-of-hooks/displayBackOfficeFooter.md
jolelievre: Actually the double `''` is the way to escape the single `'`
Or am I missing something?

kpodemski (approved):
Thank you guys @tleon @jolelievre 

I don't have any comments apart from this double `;;`, do you think you can regenerate the list @jolelievre without it?

modules/concepts/hooks/list-of-hooks/displayProductListFunctionalButtons.md
Hlavtox: @kpodemski Isn't this the hook to show buttons on product list hover that was removed “for performance reasons”?

jolelievre:
@kpodemski fixed on the core PR https://github.com/PrestaShop/PrestaShop/pull/38232 and I updated this one as well, let me know if it's all good so we can finally close this topic 😅

kpodemski (approved):

kpodemski:
merged! thank you @jolelievre @tleon ❤️

PR #1944, opened 2025-03-13
Parameters content update

Nakiloe:
The parameters content described was completely different from what is really used (came from an older version ?)

| Questions     | Answers
| Branch?       | 1.7.x / 8.x / 9.x
| Description?  | Parameters content in doc didn't match with what is really received. I've also added a note about the qtyList because it doesn't seem useful anymore (in v8.x).


kpodemski (approved):

kpodemski:
thank you @Nakiloe !

PR #1953, opened 2025-04-04
Update tutorial.md

axel-paillaud:
The logo doesn't have to be 32x32 pixels: most PrestaShop native module logos are 140x140 pixels. Also, rendering a 32x32 size is too pixelated in the back office because it's too stretched.

| Questions     | Answers
| Branch?       | 1.7.x / 8.x / 9.x


nicosomb (approved):

kpodemski:
that's true, although having it too big may display incorrectly in the back office product page if the module hooks into it, that is something to fix on the back office CSS side

so, let's remove this recommendation of 32x32 since it was mainly related to PS 1.6

PR #1982, opened 2025-05-15
Correct resource name for product_customization_field

ibobo:
The resource was missing the `product_` prefix.

| Questions     | Answers
| Branch?       | Documentation for 1.7.x, but the problem is present also in the docs for the 8.x and 9.x.
| Description?  | The page reported the resource as `customization_field` instead of `product_customization_field`.
| Fixed ticket? |


ibobo:
It seems the resource is `product_customization_fields` but the content of the document should stay `customization_field`... I think there should be a special "warning" for this.

PR #2003, opened 2025-07-09
Remove oldest informations and use v9 links

PrestaEdit:
| Questions     | Answers
| Branch?       | 9.x
| Description?  | Remove oldest informations and use v9 links.


matks (approved):
Thank you @PrestaEdit

PR #2016, opened 2025-08-06
fix: avoid $currentIndex

PrestaEdit:
| Questions         | Answers
| Branch?           | 9.x
| Description?      | Avoid $currentIndex in URLs

### DIFF
--- a/modules/creation/adding-configuration-page.md
+++ b/modules/creation/adding-configuration-page.md
@@ -105,7 +105,7 @@ public function displayForm()
     $helper->table = $this->table;
     $helper->name_controller = $this->name;
     $helper->token = Tools::getAdminTokenLite('AdminModules');
-    $helper->currentIndex = AdminController::$currentIndex . '&' . http_build_query(['configure' => $this->name]);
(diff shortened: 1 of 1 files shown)


Hlavtox (changes requested):
There should be token = false if $helper->token = Tools::getAdminTokenLite('AdminModules'); is used

modules/creation/adding-configuration-page.md
```diff
@@ -105,7 +105,7 @@ public function displayForm()
     $helper->table = $this->table;
     $helper->name_controller = $this->name;
     $helper->token = Tools::getAdminTokenLite('AdminModules');
-    $helper->currentIndex = $this->context->link->getAdminLink('AdminModules', true) . '&' . http_build_query(['configure' => $this->name]);
+    $helper->currentIndex = $this->context->link->getAdminLink('AdminModules') . '&' . http_build_query(['configure' => $this->name]);
```
Hlavtox:
```suggestion
    $helper->currentIndex = $this->context->link->getAdminLink('AdminModules', false, [], ['configure' => $this->name]);
```

Hlavtox (approved):

Touxten:
i know this bug

Touxten (approved):

PR #2025, opened 2025-08-20
fix key name

MattKelvin:
| Questions         | Answers
| Branch?           | 9.x
| Description?      | I was doing a new resource and I have discovered a small error. CQRSCommand doesn't work here it's CQRSQuery. By the way, is it not a mistake to call it CQRSQuery ? Because DeleteAttributeGroupCommand is indeed a Command. Thank you :) 
| Fixed ticket?     | /
| Sponsor company   | Mademoiselle bio

### DIFF
--- a/admin-api/contribute-to-core-api.md
+++ b/admin-api/contribute-to-core-api.md
@@ -302,7 +302,7 @@ use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
         ),
         new CQRSDelete(
             uriTemplate: '/attributes/group/{attributeGroupId}',
-            CQRSCommand: DeleteAttributeGroupCommand::class,
+            CQRSQuery: DeleteAttributeGroupCommand::class,
             scopes: [
                 'attribute_group_write',
             ],


MattKelvin:
I guess, the syntax will be good in the next version but currenlty it isn't. 

https://github.com/PrestaShop/ps_apiresources/pull/58
https://github.com/PrestaShop/PrestaShop/pull/39295/files

kpodemski:
cc @jolelievre

Touxten:
Best way is to document when the key changes ?

MattKelvin:
I think we should change it now and when the next version of PrestaShop will be published, rollback to it and explain that before it was `CQRSQuery`

admin-api/contribute-to-core-api.md
```diff
@@ -302,7 +302,7 @@ use PrestaShopBundle\ApiPlatform\Metadata\CQRSDelete;
         ),
         new CQRSDelete(
             uriTemplate: '/attributes/group/{attributeGroupId}',
-            CQRSCommand: DeleteAttributeGroupCommand::class,
+            CQRSQuery: DeleteAttributeGroupCommand::class,
```
jolelievre, line 305: No actually this was changed for 9.0.1, it was a misconception that wasn't fixed in time On the current 9.0.x branch and strating from the 9.0.1 you MUST use `CQRSCommand` which is much more in tune with the read/write separation

Besides all contribution on the module won't work if they use `CQRSQuery` because the tests are based on this recent modification in 9.0.x branch

jolelievre (changes requested):

admin-api/contribute-to-core-api.md
jolelievre: This is kind of a BC, I hesitated to keep the retro compatibility But we discussed with the other commiters from the team and reached a consensus that it's cleaner to fix this once and for good before the API gets too popular and the wrong parameter is widely used

jolelievre: By the way, it was documented:

Capture d’écran 2025-08-23 à 10 34 16

But as usual, no one reads the small lines, even the red ones 😅

MattKelvin: Thank you @jolelievre for the explanation. Sorry I miss the red lines 😄

PR #2050, opened 2025-10-08
Fix ActionColumn example to use RowActionCollection

soniiix:
The 'actions' option expects a RowActionCollection object rather than an array.

### DIFF
--- a/development/components/grid/columns-reference/action.md
+++ b/development/components/grid/columns-reference/action.md
@@ -37,7 +37,6 @@ $actionColumn->setOptions([
                 'confirm_message' => 'Delete selected item?',
             ])
         )
-     ],
 ]);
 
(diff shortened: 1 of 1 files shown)


kpodemski (changes requested):
Thanks for the pull request. I think you also have to remove `]` from line 40?

soniiix:
> Thanks for the pull request. I think you also have to remove `]` from line 40?

You're absolutely right, fixed now. My bad, I admit the heavy nesting made it a bit confusing.

soniiix:
What is happening with the progress of this PR? @kpodemski

kpodemski (approved):

kpodemski:
I forgot about it @soniiix 

thank you for this change ❤️

PR #2059, opened 2025-11-06
Update version compliancy example

Kaikina:
| Questions         | Answers
| Branch?           |  9.x
| Description?      | Example version compliancy should be updated.
| Fixed ticket?     |
| Sponsor company   | Evolutive


PululuK (dismissed):
In reality, this is not good practice, because you are blindly assuming that your module is compatible with all future versions, which may be false!

Kaikina:
> In reality, this is not good practice, because you are blindly assuming that your module is compatible with all future versions, which may be false!

That's true.

kpodemski (approved):

PR #2069, opened 2025-12-12
Clarify parameter naming in Symfony controllers notice

PrestaEdit:
| Questions         | Answers
| Branch?           | 1.7.x
| Description?      | Update parameter name


kpodemski (approved):
Thanks.

Please keep in mind that you don't need to update PrestaShop 1.7 docs. We are going to deprecate this documentation sometime soon, meaning it will stay "as is", but we won't be maintaining it.

PR #2106, opened 2026-03-04
Update URL for options.html.twig to version 8.2.x

axel-paillaud:
| Questions         | Answers
| Branch?           | 8.x / 9.x
| Description      | Branch 8.0.x doesn't exist anymore, link is broken
| Sponsor company   | Your company or customer's name goes here (if applicable).


Touxten:
Nice catch!

However, you should use the link with a tag like this so that when branch 8.2.x is deleted, the link will still be valid.

https://github.com/PrestaShop/PrestaShop/blob/8.0.0/src/PrestaShopBundle/Resources/views/Admin/Product/ProductPage/Panels/options.html.twig

Touxten (changes requested):
We should use tags, not branches, in the documentation.

axel-paillaud:
You're right, it's even better

Touxten (approved):

PR #2111, opened 2026-03-11
Update URL to point to version 8.0.0

axel-paillaud:
Branch 8.0.x doesn't exist anymore, use tag branch instead

| Questions         | Answers
| Branch?           | 8.x
| Description?      | Branch 8.0.x doesn't exist anymore, use tag branch instead
| Sponsor company   |axel-paillaud

### DIFF
--- a/modules/concepts/hooks/list-of-hooks/displayCarrierExtraContent.md
+++ b/modules/concepts/hooks/list-of-hooks/displayCarrierExtraContent.md
@@ -4,7 +4,7 @@ hidden: true
 hookTitle: 'Display additional content for a carrier (e.g pickup points)'
 files:
     -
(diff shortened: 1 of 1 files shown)


Touxten (approved):
Thanks for your PR.

What do you think about including all these changes in a single PR?
I think there must still be quite a few links to the branches?
It avoids a validation process.  

@kpodemski

kpodemski:
Hello

Yes, @Touxten is right. While we appreciate contributions, it doesn't make sense to edit one file after another in separate PRs, the URL should also point to the `9.0.x` branch, and not `8.0.0` tag.

I'm also considering using such URLs in the future:
https://github.com/search?q=repo%3APrestaShop%2FPrestaShop+displayCarrierExtraContent+language%3APHP&type=code&l=PHP

axel-paillaud:
Yes, I understand. I do it this way because I make corrections as I discover things using the documentation.

This isn't really optimal in this case.

PR #2126, opened 2026-03-25
Add PaginatedList documentation

### LINKED ISSUE #41104: Admin API doc : documenter listing via CQRS query
We have a new custom operation `CQRSPaginate` that allows having a paginated endpoint not based on a Grid factory but on a CQRS query.

It is not documented in the devdoc so we should update the doc:
- Generic doc https://devdocs.prestashop-project.org/9/admin-api/resource_server/api-resources/
- Tutorial https://devdocs.prestashop-project.org/9/admin-api/contribute-to-core-api/

Here is an example https://github.com/PrestaShop/ps_apiresources/blob/dev/src/ApiPlatform/Resources/Product/CombinationList.php#L34

tleon:
| Questions         | Answers
| Branch?           |  9.x
| Description?      | Add PaginatedList Documentation and claude skill to generate more documentation on custom endpoints
| Fixed ticket?     | ~
| Sponsor company   | PrestaShop SA

### DIFF
--- a/.claude/document-custom-operation/SKILL.md
+++ b/.claude/document-custom-operation/SKILL.md
@@ -1,0 +1,68 @@
+Generate developer documentation for a PrestaShop Admin API custom operation class.
+
+**Arguments:** `$0` = operation name, `$1` = PrestaShop core path
+
+Example invocation arguments: `toggle-exchange-rate-automatization-command /home/tleone/Dev/prestashop-develop`
+
+## Task
+
+You are writing documentation for the PrestaShop devdocs site (https://devdocs.prestashop-project.org/9/admin-api/resource_server/api-resources/). The target page is `admin-api/resource_server/api-resources.md` in this repository. Your output will be a new section to add to that file for operation `$0`.
+
+## Steps
+
+### 1. Read the operation class
+
+The PrestaShop core source is at `$1`. Find operation `$0` in `src/PrestaShopBundle/ApiPlatform/Metadata/`. Read it fully to understand:
+- What it extends and what interfaces it implements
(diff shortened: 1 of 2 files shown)


.claude/document-custom-operation/SKILL.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Task
+
+You are writing documentation for the PrestaShop devdocs site (https://devdocs.prestashop-project.org/9/admin-api/resource_server/api-resources/). The target page is `admin-api/resource_server/api-resources.md` in this repository. Your output will be a new section to add to that file for the given operation.
+
+## Steps
+
+### 1. Read the operation class
+
+The PrestaShop core source is at `/home/tleone/Dev/prestashop-develop`. Find the class in `src/PrestaShopBundle/ApiPlatform/Metadata/`. Read it fully to understand:
```
kpodemski: You should replace it with variable. https://code.claude.com/docs/en/skills#available-string-substitutions

kpodemski (changes requested):
To be honest, I'm not entirely sure if this skill is relevant to handling updates on just one page of the documentation, this way 🤔

tleon:
This is mostly the first time I was using a skill for claude and it was mostly testing around the skill but in the end I thought what's the point of not committing it. If in the future we introduce some more custom operation well we'll be able to document them using this skill.

jolelievre (approved):
I agree the skill is a bit overkill but it doesn't hurt either since it's here now

We're only at the beginning of using AI, so maybe some things won't be relevant but it's worth testing stuff, and we can always clean them later when we'll be more aware of what is relevant or not

kpodemski (approved):

